PR #8, opened 2020-07-05
feature/event-card

### LINKED ISSUE #3: Event card
Related to  [#79](https://github.com/CouncilDataProject/seattle/issues/79).
Write an event card component and update the styling to match the UX team's designs.

trungvngo:

Resolves #3 

Work in progress: Added a `Base` card component and a `MeetingCard` component. 
Screen Shot 2020-07-05 at 3 14 41 PM

Open questions:
- [x] Should the entire meeting card be a clickable element that will take you to the meeting page should only certain elements be clickable?
- [x] What format will "actions" and "tags" come in? e.g. tags will be a comma separate list and actions is a string that the component can render without further manipulation


trungvngo:
@tohuynh @JacksonMaxfield Some open questions in the PR description I'd love your feedback on.

tohuynh:
> @tohuynh @JacksonMaxfield Some open questions in the PR description I'd love your feedback on.

1. I'm OK with making the entire card clickable or just the image and committee name only.
2. "actions" and "tags" are generated from [event schema's](https://github.com/CouncilDataProject/cdptools/blob/release-3/docs/document_store_schema.md) `matters` and `keywords`, respectively. I think I prefer passing the list of matters and keywords as props to the event card component and generating those strings inside the event card component.

JacksonMaxfield:
First: Thanks for getting this started!
Second: I hate to do this to you but the designs have been slightly updated after feedback from user testing. But not _much_ different.

Mobile Event Card:
image

Desktop Event Card:
image


Answering questions:
1. I believe the intention from UX is that the entire card is clickable (and personally that makes the most sense to me)
2. Fortunately with the new designs you don't need to receive or display the `matters`. Otherwise agree with @tohuynh as basically assume you will receive the `Event` object from the new schema as params.

trungvngo:
@JacksonMaxfield No worries about the updated designs! Can you share a link to the designs in the context of the meetings and search page? It would be helpful to understand the layout. Also, any additional details on dimensions of the cards and font sizes would be helpful if it's available.

JacksonMaxfield:
> @JacksonMaxfield No worries about the updated designs! Can you share a link to the designs in the context of the meetings and search page? It would be helpful to understand the layout. Also, any additional details on dimensions of the cards and font sizes would be helpful if it's available.

@trungvngo I don't actually have those. Want to ping Emily in the UX-Design channel on slack?

trungvngo:
@JacksonMaxfield @tohuynh I updated to match Emily's latest designs. You can test the component out in Storybook. I ended up not using the SemanticUI Card component. It really wasn't doing much, and was more of a hassle to override styles than to just build the component from scratch. I also didn't make it clickable at the component level. Since we want the entire card to be clickable the consumer can just wrap the component in a link. That way this component isn't opinionated about routing.

src/components/Cards/MeetingCard.tsx
```diff
@@ -0,0 +1,87 @@
...
+  font-weight: bold;
+  color: ${colors.black};
+`;
+
+const SearchExcerpt = styled.div`
+  font-size: ${fontSizes.font_size_3};
+  color: ${colors.grey};
+`;
+
+export const MeetingCard = (props: MeetingCardProps) => {
```
tohuynh, line 47: It is not needed to typed MeetingCard with `FunctionComponent<MeetingCardProps>` as long as the props is typed?

src/components/Cards/MeetingCard.tsx
```diff
@@ -0,0 +1,87 @@
+import React from "react";
+import styled from "@emotion/styled";
+import { Card } from "./";
+import colors from "../../styles/colors";
+import { fontSizes } from "../../styles/fonts";
+
+type MeetingCardProps = {
+  imgSrc: string;
```
tohuynh, line 8: @JacksonMaxfield  Are we going to also add thumbnail_hover_file (animated gif of the meeting, when the user hovers over the card) to the meeting card?

src/components/Cards/Base.tsx
```diff
@@ -0,0 +1,76 @@
...
+  position: relative;
+  /* set min-height for image to prevent div collapsing in small screens */
+  min-height: ${IMAGE_CONTAINER_MIN_HEIGHT};
+  border-radius: ${CARD_BORDER_RADIUS} ${CARD_BORDER_RADIUS} 0 0;
+  background-image: linear-gradient(
+      transparent 50%,
+      rgba(51, 51, 51, 0.1) 60%,
+      rgba(51, 51, 51, 0.3) 70%,
+      rgba(51, 51, 51, 0.5) 80%,
+      rgba(51, 51, 51, 0.8) 100%
```
tohuynh, line 37: If I'm understanding this correctly this makes container gradually darker to ensure there's enough contrast between the white date text and the background? Is the white date the source of the accessibility 'incomplete' warning on storybook?

src/components/Cards/MeetingCard.tsx
trungvngo: If I understand your question correctly, yes! The component gets its type from the props:
Screen Shot 2020-09-19 at 3 42 23 PM

src/components/Cards/Base.tsx
trungvngo: That's correct. This adds the gradient. Also correct that it is the source of the accessibility incomplete warning. Any thoughts on how to address this warning?
Screen Shot 2020-09-19 at 3 50 09 PM

trungvngo:
@JacksonMaxfield @tohuynh sorry for the delay! I've made the requested changes. Let me know what you think!

src/components/Cards/Base.tsx
JacksonMaxfield: IIRC, you can have both a `background-color` and a `background-image` property on an element and `background-color` is "underneath" the image. In this case it may be good to just default to something that we know will contrast from the text as the color as a safety / default?

JacksonMaxfield (approved):
Looks good to me. I think the`background-color` thing is fine but would love @tohuynh's opinion.

JacksonMaxfield (changes requested):
Any chance to fix the broken build? I tried my hand at it but it's some TypeScript handling that is a bit above me. I could probably toss more hours at it if needed.

tohuynh:
> Any chance to fix the broken build? I tried my hand at it but it's some TypeScript handling that is a bit above me. I could probably toss more hours at it if needed.
@JacksonMaxfield I believe adding typing to the styled component will do the trick.
`const CardImageContainer = styled.div<CardImageContainerProps>` to line 26 of Base.tsx.

src/components/Cards/Base.tsx
tohuynh: If we were to add a background-color, it would have to be in the `span` element where the violation occurs (line 54-61 of MeetingCard.tsx). Because if we had a background-color to `CardImageContainer` the violation would still occur. Not sure what to do.

JacksonMaxfield: I think because it is "incomplete" and not a "violation" I am okay with merging it in. We can think about how to solve it later.

JacksonMaxfield: If we come up with a better solution that is. 🤷

JacksonMaxfield:
> @JacksonMaxfield I believe adding typing to the styled component will do the trick.
> `const CardImageContainer = styled.div<CardImageContainerProps>` to line 26 of Base.tsx.

Thanks @tohuynh, that did the trick.

JacksonMaxfield:
@trungvngo @tohuynh I just pushed changes to fix the build. With that, I am happy with the changes. Sorry this took so long to resolve. I think we all got rather busy / distracted. Thanks for all the help!

Will merge once the checks pass 🎉

JacksonMaxfield (approved):

trungvngo:
Thank you both for fixing the build. Great to see this merged in!

Issue #12, opened 2020-11-13
Update License to MIT


JacksonMaxfield:
`cdp-backend` went with MIT license. Let's just change them all over to MIT.

JacksonMaxfield:
Resolved by #14

PR #28, opened 2021-02-11
feature/adopt-mozilla-protocol

### LINKED ISSUE #11: Adopt Mozilla Protocol wherever possible
## Use Case

Offload some of the design and styling maintainence work while also having nice standardize, accessible design.

## Solution

[Mozilla protocol](https://protocol.mozilla.org/) is a really good solution. Both because I have long been a fan of Mozilla styling but also because of how easy it is to use (imo).

This doesn't mean scrap all the UX design work already done. I think we should simply merge the two.

For example, instead of writing our own cards, we can simply use [theirs](https://protocol.mozilla.org/patterns/molecules/card.html#extra-small-card) but add all the extra stuff we need. I.E. tags, etc.

Also see: [Mozilla branding style guide](https://mozilla.design/mozilla/) for great examples and quick rundown of their choices.

## Alternatives

hawkticehurst:
- [x] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Resolves #11

A close review of the changes (especially regarding the FilterPopup component) would be appreciated as I think there are probably some issues that still need to be resolved since I couldn't use 100% Mozilla styling for it.

Also @tohuynh, I believe we have to coordinate the changes we are both making to the Card component. I'm happy to copy whatever you're currently doing if you'd like––just let me know.


src/components/Cards/MeetingCard/MeetingCard.tsx
```diff
@@ -1,87 +1,53 @@
...
+const MeetingCard = ({
+  cardLink,
+  imgSrc,
+  imgAlt,
+  imgOverlayText,
+  committee,
+  tags,
+  excerpt,
+}: MeetingCardProps) => {
+  const tagString = tags.join(` • `);
```
JacksonMaxfield, line 34: Just curious, is there an html character code to use instead of a unicode character?

src/components/Cards/MeetingCard/MeetingCard.tsx
```diff
@@ -1,87 +1,53 @@
...
+
+const DateOverlay = styled.p`
+  position: absolute;
+  z-index: 1;
+  color: white;
+  font-size: 30px;
+  font-weight: bolder;
+  top: 75%;
+  left: 2%;
+`;
```
JacksonMaxfield, line 13: There is a part of me that thinks maybe we shouldn't have the date overlay on the image and just have it as content on the card... I mean we are getting an accessibility warning for a reason...

src/components/Filters/FilterPopup/FilterPopup.tsx
```diff
@@ -31,22 +40,50 @@ const ContentContainer = styled(Popup.Content)({
...
+const MozillaNeutralButton = styled.button({
+  color: "#5e5e72",
+  backgroundColor: "transparent",
+  border: "2px solid #000000",
+  borderColor: "#cdcdd4",
+  borderRadius: "4px",
+  padding: "6px 24px",
+  fontSize: "0.875rem",
+});
+MozillaNeutralButton.displayName = "MozillaNeutralButton";
```
JacksonMaxfield, line 78: Sucks that we have to do this but :shrug:

JacksonMaxfield (commented):
This all makes sense to me. Not approving yet as failing builds but when those are fixed all good.

For the sake of documentation in this PR, can you add screenshots to the PR of the new cards?

src/components/Cards/MeetingCard/MeetingCard.tsx
```diff
@@ -1,87 +1,53 @@
...
+        <div className="mzp-c-card-media-wrapper">
+          <DateOverlay>{imgOverlayText}</DateOverlay>
+          <img className="mzp-c-card-image" src={imgSrc} alt={imgAlt} />
+        </div>
+        <div className="mzp-c-card-content">
+          <div className="mzp-c-card-tag">Committee</div>
+          <h2 className="mzp-c-card-title">{committee}</h2>
+          {excerpt ? <p className="mzp-c-card-desc">{`"${excerpt}"`}</p> : null}
+          <p className="mzp-c-card-meta">Tags: {tagString}</p>
+        </div>
```
tohuynh, line 48: Here's what I used to display the the tags:
```
<p className="mzp-c-card-meta">Tags</p>
<p className="mzp-c-card-desc">{tagString}</p>
```

tohuynh (commented):
I think the build is failing due to cdp-frontend's webpack config doesn't have a loader for the Mozilla protocol css file. I'm not entirely sure how to fix it though.

src/components/Cards/MeetingCard/MeetingCard.tsx
tohuynh: There is `&bull;`, but we'd have to use [dangerouslysetinnerhtml](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) to create the markup for tagString. Which is fine I think.

src/components/Cards/MeetingCard/MeetingCard.tsx
tohuynh: Also I changed <a className="mzp-c-card-block-link" href={cardLink}> to a <div> element and removed href accordingly. The reason was to keep the card unopinionated about routing to the card details page.  The consumer of the card can always wrap it in a link.

src/components/Cards/MeetingCard/MeetingCard.tsx
tohuynh: Yeah, the warning just means it couldn't figure out if there's enough color contrast, but manually checking there is enough color contrast. I like getting rid of the overlay too. The only advantage I could think of for keeping the overlay is creating more room for the other content.

JacksonMaxfield: Agree but I think we can make the cards larger if we need more space AND we can fix the accessibility warning. Which imo, the accessibility warning is important to address.

JacksonMaxfield:
Hey @hawkticehurst any chance we could get this merged by March 13th (the next hackathon)? Would be nice to have all the components looks the same when explaining how we are planning to work.

src/components/Cards/MeetingCard/MeetingCard.tsx
hawkticehurst: I vote we don't use dangerouslySetInnerHTML too 😅. So I'm going to go ahead and resolve this conversation.

src/components/Cards/MeetingCard/MeetingCard.tsx
hawkticehurst: Another option would be to simply add a background to the overlay, thus saving room and guaranteeing accessibility.

hawkticehurst: This is a very quick change so I'll push something up so you guys can give feedback on the design.

src/components/Filters/FilterPopup/FilterPopup.tsx
hawkticehurst: Agreed :( I still feel icky about combining semantic UI with the rest of the library. But short of re-implementing the whole component from scratch, I think this is about as good as we're going to get for now.

And for the record, I think reimplementing this component is something worth looking into later but for the sake of finishing up the adoption of Mozilla styling, we'll just have to leave it.

JacksonMaxfield (approved):
THANK YOU! :heart:

Issue #21, opened 2020-12-11
Adopt Storybook Controls Addon


hawkticehurst:
## Background
Per this [FAQ](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#faqs) Storybook will be deprecating the Knobs addon by v7.0 and will prefer the usage of Controls addon.

## Solution
Follow this [transition guide](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs) to update all stories that use Knobs to instead use Controls.

Additionally, the [docs page](https://storybook.js.org/docs/react/essentials/controls) for Controls can be referenced for more information.

We also should likely update our own contribution docs to make sure that the continued use/preference of Controls for future stories is made clear.

## Alternatives
Leave Knobs as is and just make sure not to update Storybook to < v7.0. Since the transition to Controls does not seem to be that challenging, however, I think this alternative would be ill-advised.

PR #30, opened 2021-02-27
feature/add-person-card

### LINKED ISSUE #5: Council member card
Related to  [#79](https://github.com/CouncilDataProject/seattle/issues/79).
Write a council member card component and update the styling to match the UX team's designs.

tohuynh:
  Resolves #5 
  Add person card. [Preview](https://tohuynh.github.io/cdp-frontend/?path=/story/library-cards-person--default-story)

Thanks for contributing!

Issues to discuss:
- How to indicate that person is active? The UX team's recommendation is to remove the map when the person is not active. That makes sense to me to say that person doesn't have a spot on the map if they are not currently in office. But is there a better way to indicate the person is in office? Something that requires less thinking from the user. An icon maybe?
- For `Chair`, is that meant to display one or many committees?


JacksonMaxfield:
This looks great!

Answering questions:

> How to indicate that person is active? The UX team's recommendation is to remove the map when the person is not active. That makes sense to me to say that person doesn't have a spot on the map if they are not currently in office. But is there a better way to indicate the person is in office? Something that requires less thinking from the user. An icon maybe?

Great question, more iconography is, I think, generally a good thing. I wouldn't be opposed to finding an icon for active or inactive member that we can float right on the same line as the name? (or at least have similar positioning with the legislation card?). Especially because I am not sure if I like the map anymore.... / it's a hard requirement for instances to generate map images. Can you make the map part optional and if the map isn't provided the person image takes over the whole top of the card?

> For Chair, is that meant to display one or many committees?

Another great question. I think in Seattle each member only chair's one committee but that may not be true in other places.... Does the database schema even allow for chairing multiple committees? -- Answering my own question, it does: [seat + role docs](https://councildataproject.github.io/cdp-backend/cdp_backend.database.html#cdp_backend.database.models.Role) and [filled ingestion models example](https://councildataproject.github.io/cdp-backend/ingestion_models.html#expanded-event-data-with-event-minutes-items) shows we can have multiple "Chair" and "Vice Chair" for a single person in which case I would make that a list of roles or maybe just the "highest priority role" and then say like "(and two other roles)"? Would love @hawkticehurst opinion on this one and I may reach out to Emily to ask her opinion too.

hawkticehurst:
Agreed, this looks great!

To answer the in-office versus out-of-office question, I'm not sure enabling and disabling a map would immediately be clear that it indicates in-office versus not and thus think it's probably not a great UX flow.

An alternative idea would be to use a sort of badge component. I could imagine a green (or maybe grey) badge with the text "Active Member" or "In Office" overlayed on the bottom left corner of the image. Likewise out of office members could have a red (or grey) badge with relevant text or even no badge at all.

In this way, the badge would still be accessible (light text over solid dark background), but we could still place the badge over the image and save space.

Unfortunately, I can find a badge component on Mozilla Protocol but I'm very confident it would not be hard to create a custom badge in the likeness of that design system.

JacksonMaxfield:
> An alternative idea would be to use a sort of badge component. I could imagine a green (or maybe grey) badge with the text "Active Member" or "In Office" overlayed on the bottom left corner of the image. Likewise out of office members could have a red (or grey) badge with relevant text or even no badge at all.

I like this idea. I also wonder if instead of the map we just have another text component that has _in text_ the neighborhoods / districts they represent. I.e. for Sawant it would say "District 3 -- Capitol Hill, First Hill, Madison Valley" and for Mosqueda it would say "Position 8 -- All City". I really am just not convinced that the images are a good strategy for both accessibility + scalablity. (As much as I love them. I think you know how hyped I was about them)

tohuynh:
>Can you make the map part optional and if the map isn't provided the person image takes over the whole top of the card?

Yes

>I also wonder if instead of the map we just have another text component that has in text the neighborhoods / districts they represent. I.e. for Sawant it would say "District 3 -- Capitol Hill, First Hill, Madison Valley" and for Mosqueda it would say "Position 8 -- All City".

We already have the electoral area they represent in the card though.

>I wouldn't be opposed to finding an icon for active or inactive member that we can float right on the same line as the name? (or at least have similar positioning with the legislation card?).

I couldn't think any of active/inactive icons other than person figure with check(active) icon and person figure with x(inactive). This is probably not that obvious to the user what these icons should represent.

What about Lisa Herbold (status), where the status is in smaller text?

>An alternative idea would be to use a sort of badge component. I could imagine a green (or maybe grey) badge with the text "Active Member" or "In Office" overlayed on the bottom left corner of the image. Likewise out of office members could have a red (or grey) badge with relevant text or even no badge at all.

@hawkticehurst Did you already create something similar while working the date overlay for the meeting card? We can use that if yes.

JacksonMaxfield:
> We already have the electoral area they represent in the care though.

Wow, can't believe I forgot about that after looking at the card 10 seconds eariler. My bad. Yes. I guess my recommendation is to simply have the map be optionally visible based off parameter provided or not.

Thinking about this further, whenever we get to making the "Person Card Container" thinking about the visual design of it all, I would say that unless the database has links to maps for all people, only then will the maps render. I think it would look visually odd if 7 / 9 of the people had maps and 2 didn't for example. So the container will have to do some logic about providing the map value to each card or not. (This is future work obviously not on this PR), if you agree @tohuynh I will make a GH Issue for that just to track.

> I couldn't think of any active/inactive icons other than person figure with check(active) icon and person figure with x(inactive). This is probably not that obvious to the user what these icons should represent.
> What about Lisa Herbold (status), where the status is in smaller text?
> @hawkticehurst Did you already create something similar while working on the date overlay for the meeting card? We can use that if yes.

Linking all of these comments together because after thinking about it I think I know what _may_ be nice. Instead of an icon we could do something like the GitHub "labels":
image

Were we float right a just colored element that is "active" or "in-active"?

JacksonMaxfield:
I will also say, I am not entirely sold on the "labels" idea. I think it would conflict with the rest of the visual design of the cards / the legislative card having icongraphy and this card having labels.... @tohuynh do you have a link to the icon site we were using? I can look around.

tohuynh:
>@tohuynh do you have a link to the icon site we were using? I can look around.
I only look through https://heroicons.com/
But https://react-icons.github.io/react-icons/ has lots more.

tohuynh:
Yeah, too bad protocol doesn't have a labels component

tohuynh:
>Thinking about this further, whenever we get to making the "Person Card Container" thinking about the visual design of it all, I would say that unless the database has links to maps for all people, only then will the maps render. I think it would look visually odd if 7 / 9 of the people had maps and 2 didn't for example. So the container will have to do some logic about providing the map value to each card or not. (This is future work obviously not on this PR), if you agree @tohuynh I will make a GH Issue for that just to track.

Agree

JacksonMaxfield:
I am tempted to say that we could use `user-remove` with a light grey color and `user-group` in some cdp color for inactive and active respectively.

The idea in my head being, "they are no longer in the group" :shrug:

tohuynh:
https://react.semantic-ui.com/elements/label/#types-ribbon

Do you like this ribbon idea?

JacksonMaxfield:
> https://react.semantic-ui.com/elements/label/#types-ribbon
> 
> Do you like this ribbon idea?

Nah not too much personally.

hawkticehurst:
> @hawkticehurst Did you already create something similar while working on the date overlay for the meeting card? We can use that if yes.

I kind of did but ultimately deleted it sadly. That being said I did a very simple case of creating a custom element using Emotion and then added styling that included:

- Absolute positioning so the badge was overlayed over the photo
- Usage of `top` and `bottom` to get the badge in the correct spot (also used percentages as the unit)
- Some easy font changes
- Border radius
- Some light padding

tohuynh:
Regarding the active/inactive visual cue, I went with a hybrid approach. Use the same positioning as the status of the legislation card, but use labels instead of icons. That way we can replace the labels with better icons later on. While `user-group` and `user-remove` make sense to me, I don't think they are obvious enough to users.  I ended up using `active`(in cdp-neon-green) and `inactive`(in cdp-dark-grey) labels

Also used storybook controls addon in the stories to help with #21. 
[New preview](https://tohuynh.github.io/cdp-frontend/?path=/story/library-cards-person--with-seat-picture)

hawkticehurst:
Changes look good to me!

hawkticehurst (approved):

PR #41, opened 2021-03-15
feature/minutes-item-card

andrewlubrino:
**Pull request recommendations:**
- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [x] Link to any relevant issue in the PR description. Ex: _Resolves [gh-12], adds tiff file format support_

Resolves #35 

- [x] Provide context of changes.

Adds the Minutes Item Component where you can accordian view attachments to each minutes item.

- [ ] Provide relevant tests for your feature or bug fix.
- [x] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

Issue #35, opened 2021-03-11
Create Minutes Item Component


JacksonMaxfield:
On the event page, one of the tabs is the Minutes + Documents.

This will need to be broken down into a container + a minutes item component.

![image](https://user-images.githubusercontent.com/17132317/110734496-cacaab00-81dc-11eb-98db-182aa0bfb965.png)

A minutes item component will be each individual item in the list. It should have the functionality to expand and see the related documents to that item.

JacksonMaxfield:
Closed in #42

PR #42, opened 2021-03-23
feature/minutes-item

andrewlubrino:
- Links to relevant issues: #41 & #35 
- Made summary text black / underlined, document links list made flush with text above, uses emotion/styled. Various other small changes requested by team.
- Updated minutes item with requested changes from #41 .
- [Link to GitHub Pages](https://andrewlubrino.github.io/cdp-frontend/?path=/story/library-details-minutesitem--minutes-item)

I ended up taking out the `<li>` elements from the map function in the `returnDocsList` function (see lines 42 through 46 in `MinutesItem.tsx`) because I was having trouble using emotion/styled to remove the indentation. If this causes any issues, let me know and I can fiddle around with it some more.


src/components/Details/MinutesItem/MinutesItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface Items {
+  /*Minute item headline */
+  item: string;
+  /*Array of documents to be shown below*/
+  docs?: Docs[];
+}
+
+export interface MinutesItemProps {
+  minutesItems: Items[];
+}
```
tohuynh, line 30: Could add a comment for this props field as well?

tohuynh (approved):
Looks great! Thanks for making the changes.

Regarding removing list indentation, did you already try fiddling with [list-style-position](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position)?

src/components/Details/MinutesItem/MinutesItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <div>{elem.item}</div>
+            <div>{returnDocList(elem.docs)}</div>
+          </li>
+        );
+      })}
+    </ol>
+  );
+};
+
+export default MinutesItem;
```
JacksonMaxfield, line 68: My nitpick is to ask to rename the prop from `minutesItems` to `attachments`? Because the number list of this component is the full "Minutes Item List" where each minutes item _may have_ some attachments.

JacksonMaxfield (approved):
This looks good to me! One minor nitpick and one question for all other reviewers.

Anyone else have the "See documents" split across two lines? Can we make that a single line?

JacksonMaxfield:
image

^ Image of the text split across two lines

src/components/Details/MinutesItem/MinutesItem.tsx
tohuynh: I think you mean change `Docs` to `attachments`, right?

JacksonMaxfield: uhhhhhhhh woops. Alright it's not the prop that is confusing me its the name of the component. Can we name it "MinutesItemsList"? A single MinuteItem isn't what we are doing here, this is the full list right?

tohuynh: Yes, that's right, this is the full list. So yes, good call on MinutesItemsList

src/components/Details/MinutesItem/MinutesItem.tsx
andrewlubrino: Yes, I tried `listStylePosition: "outside"` and it didn't change the indentation.

andrewlubrino:
> Looks great! Thanks for making the changes.
> 
> Regarding removing list indentation, did you already try fiddling with [list-style-position](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position)?

Yes, I tried `listStylePosition: "outside"` and that didn't change the indentation.

andrewlubrino:
> image
> 
> ^ Image of the text split across two lines

That's super weird. I wonder why that's happening. I bumped up the width of the container for the summary/details elements. Hopefully that works across the board.

andrewlubrino:
Unless there are further comments tonight, I'll make another PR either sometime late tonight or tomorrow.

JacksonMaxfield:
> Unless there are further comments tonight, I'll make another PR either sometime late tonight or tomorrow.

@andrewlubrino No need to make a new PR. If you just keep working on the same branch and push commits this PR will update with the new commits!

andrewlubrino:
> > Unless there are further comments tonight, I'll make another PR either sometime late tonight or tomorrow.
> 
> @andrewlubrino No need to make a new PR. If you just keep working on the same branch and push commits this PR will update with the new commits!

That makes a lot of sense. I'll make sure to just work on the same branch in the future instead of starting new PRs. Thanks! I'll send some changes through shortly then.

JacksonMaxfield:
@andrewlubrino can you update your gh-pages / storybook when you get a chance as well?

andrewlubrino:
> @andrewlubrino can you update your gh-pages / storybook when you get a chance as well?

Okay, just deployed to gh-pages again. Is "See documents" still on two separate lines for you?

JacksonMaxfield:
Hm. It is still on two lines for me. I am on firefox. But this has led me to maybe change my mind. I think the plus / x icon are floated right to the edge of the container without width specified, is that correct? If so, then I think the better option (and this is going back on my prior statement so apologies), would be to simply control the container width of this whole minutes items component from whatever container we wrap this in.

Maybe @tohuynh can chime in. This component would be wrapped by the "EventDetails" container right? In the larger view of this component I think it would make sense to just have that icon float right to the size of the container.

This is the one downside of storybook I have found is that the storybook canvas width is the same width as the whole container which may not be true of the final app.

JacksonMaxfield:
I played around with the css a bit. If we do want to enforce width on this specific icon, `10em` works for me.

JacksonMaxfield:
But my most recent question should be answered first probably which is a larger discussion.

JacksonMaxfield:
Sorry @andrewlubrino for going back and forth on something so minor.

JacksonMaxfield:
Thanks for all the changes @andrewlubrino. Discussed during dev meeting. Merging :slightly_smiling_face:

PR #44, opened 2021-04-08
feature/create-exportable-app

### LINKED ISSUE #13: Create exportable App to import into cookiecutter
## Use Case

Get started on the full exportable app that this library is ultimately responsible for.

## Solution

Thinking out loud:
A app that a cookiecutter instance can import from the library. Notably I have two initial ideas as to what it may be good to parametrize but willing to hear other thoughts.

1. `instance.css` a css file that contains color definitions, i.e. `primary`, `secondary`, `underline`, `highlight`, etc. Not sure how many colors we will have customizable in the app itself but just the general idea that some of the styling in an instance is customizable.
2. `instance.json` a JSON file that stores metadata for the web app to use, i.e. links to the official government sources, etc. Again, not sure how much we will need in this config but there is an assumption that it will be used.

Ultimately it would great if the end result of this specific issue was for the [cookiecutter-cdp-deployment](https://github.com/councildataproject/cookiecutter-cdp-deployment) to import the `CDPApp` (or w.e. name we go with) into it and for that app to render a header, general landing page, and footer. The footer content can be filled with the content of the `instance.json` and maybe some of the content on the landing page can be colored using the `instance.css`.

## Additional Notes
Please keep in mind the desire to move to Mozilla styling as per #11 but here is the mock landing page from UX team:
image

This was inspired from [ZocDoc](https://www.zocdoc.com/), but would again would _personally_ prefer and side on a mix of this and Mozilla styling.

## Alternatives

hawkticehurst:

Resolves #13.

Implements and exports an initial static `App.tsx` that can be used in cookiecutter.

To help with the future development of the CDP app, `snowpack` was installed and an `npm start` script was created that will start up a development server.

- [ ] Provide or update documentation for any feature added by your pull request.

@tohuynh @JacksonMaxfield You can install the changes and run `npm start` to run a dev server locally. Also, this was tossed together very quickly, so feel free to ask for any changes you want.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!-- 
-  This is file is used as the entry point for the Snowpack development server.
+  This is the file is used as the entry point for the Snowpack development server.
   This file must stay in the root directory so it can be detected by Snowpack.
  -->
 <html lang="en">
--- a/index.html
+++ b/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Document</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="./src/index-react.js"></script>
+  </body>
+</html>
--- a/index.html
+++ b/index.html
@@ -1,3 +1,7 @@
+<!-- 
+  This is file is used as the entry point for the Snowpack development server.
+  This file must stay in the root directory so it can be detected by Snowpack.
+ -->
 <!DOCTYPE html>
 <html lang="en">
   <head>
--- a/index.html
+++ b/index.html
@@ -1,8 +1,8 @@
+<!DOCTYPE html>
 <!-- 
   This is file is used as the entry point for the Snowpack development server.
   This file must stay in the root directory so it can be detected by Snowpack.
  -->
-<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@
     <meta charset="UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Document</title>
+    <title>Council Data Project</title>
   </head>
   <body>
     <div id="root"></div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,0 +1,79 @@
+import React from "react";
+import { Footer } from "./index";
+import "@mozilla-protocol/core/protocol/css/protocol.css";
+
+function App() {
+  return (
+    <>
+      <header
+        style={{
+          display: "flex",
+          justifyContent: "space-between",
+          alignItems: "center",
+          padding: "1rem",
+          borderBottom: "solid 1px rgba(0,0,0,0.2)",
+        }}
+      >
+        <h1 style={{ fontSize: "25px", margin: 0 }}>Council Data Project</h1>
+        <section>
+          <a href="#" style={{ padding: "1rem" }}>
+            Events
+          </a>
+          <a href="#" style={{ padding: "1rem" }}>
+            People
+          </a>
+        </section>
+      </header>
+      <main
+        style={{
+          display: "flex",
+          flexFlow: "column wrap",
+          justifyContent: "center",
+          alignItems: "center",
+          height: "85vh",
+        }}
+      >
+        <h1 style={{ fontSize: "30px", marginBottom: "2rem" }}>Search City Council Transcripts</h1>
+        <section>
+          <input type="text" style={{ marginRight: "1rem", width: "30vw" }}></input>
+          <button
+            style={{
+              border: "2px solid transparent",
+              color: "#ffffff",
+              fontSize: "1rem",
+              fontWeight: "bold",
+              background: "#0060df",
+              padding: "8px 24px",
+              borderRadius: "4px",
+            }}
+          >
+            Search
+          </button>
+        </section>
+      </main>
+      <footer>
+        <Footer
+          footerLinksSections={[
+            {
+              footerLinksSectionName: "Seattle",
+              links: [
+                { label: "City Of Seattle", url: "#" },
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -0,0 +1,79 @@
...
+                { label: "Seattle Channel", url: "#" },
+              ],
+            },
+            {
+              footerLinksSectionName: "Developers",
+              links: [
+                { label: "Seattle Specific", url: "#" },
+                { label: "Transcription Runner", url: "#" },
+              ],
+            },
```
JacksonMaxfield, line 71: Nit: remove these old archaeic references... Just remove this section entirely.

But for a bit of fun, if you want to see what they used to point to:
https://github.com/CouncilDataProject/seattle_v1
https://github.com/CouncilDataProject/cdptools_v1

src/App.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <Footer
+          footerLinksSections={[
+            {
+              footerLinksSectionName: "Seattle",
+              links: [
+                { label: "City Of Seattle", url: "#" },
+                { label: "Seattle Open Data", url: "#" },
+                { label: "Seattle Channel", url: "#" },
+              ],
+            },
```
JacksonMaxfield, line 64: Nit: just make these "Lorem", "Ipsum" and the section be "Links"

index.html
```diff
@@ -0,0 +1,17 @@
...
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Council Data Project</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="./src/index-react.js"></script>
+  </body>
+</html>
```
JacksonMaxfield, line 17: For my clarification, the cookiecutter will also need this file + SEO stuff in here right?

https://github.com/CouncilDataProject/cookiecutter-cdp-deployment/blob/main/%7B%7B%20cookiecutter.hosting_github_repo_name%20%7D%7D/web/public/index.html

src/index.ts
```diff
@@ -1,3 +1,6 @@
+// Default App
+export { default as App } from "./App";
```
JacksonMaxfield, line 2: I thought about this last night. If we make this a default export, and someone wants to _only_ use one or two of our components, will they be forced to import the full App as well (thereby importing all the components)?

JacksonMaxfield (approved):
Minor nitpicks but:

S O  E X C I T E D

src/App.tsx
hawkticehurst: As a future thought for the Footer actually, I think we should allow the link sections to be optional because right now it's required to pass something in.

index.html
hawkticehurst: Yes, on the cookie-cutter side of things you will need to create an `index.html` and something similar to the `index-react.tsx` file so the `App` component can be rendered. 

What might differ is where the `index.html` file lives within the project directory depending on what tooling we use for a dev server and to build the app.

For example, I imagine one of the easiest routes would actually be to use the `create-react-app` script in cookie-cutter and then just delete all the boilerplate we don't need and import this version of `App`.

This is also certainly something @tohuynh should chime in on.

src/App.tsx
JacksonMaxfield: Yea that makes sense to me. Didn't realize that.

src/index.ts
hawkticehurst: If we make this a default export, users will have the flexibility to import whatever they want. So someone could choose to write `import {Footer, MeetingCard} from @cdp/frontend;` and then the `App` would never be downloaded.

Comparitively, an install with `App` and the components would look like so, `import App, {Footer, MeetingCard} from @cdp/frontend;`.

Nowwww with all that said, the line you highlighted is actually a named export for App (the curly braces mean named export) because I couldn't quite figure out how to make App a default export from src/index.ts and wanted to be able to move on and work on `App`. So in reality someone importing will write this:

`import {App} from @cdp/frontend;`

Or this if they want more components:

`import {App, Footer, MeetingCard} from @cdp/frontend;`

tohuynh (approved):

src/index.ts
JacksonMaxfield: Weird. I think I am partial to making it a non-default export for some reason.

Crazy that you can have two export patterns in a single line.

index.html
JacksonMaxfield: I guess we could submodule the cookiecutter repo into this repo?

JacksonMaxfield: I don't think syncing them is a priority for now though. A nice to have for sure but I would add an issue for it to track / keep us aware that we need to sync them eventually.

src/App.tsx
```diff
@@ -0,0 +1,79 @@
...
+              padding: "8px 24px",
+              borderRadius: "4px",
+            }}
+          >
+            Search
+          </button>
+        </section>
+      </main>
+      <footer>
+        <Footer
```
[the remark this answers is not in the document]
hawkticehurst: Whoops, that was an accident. Thanks for the catch.

index.html
hawkticehurst: Yeah, I don't think syncing is super necessary either. The main place we need to get those details correct is just in cookie-cutter since that is where the final form of `index.html` and `index-react.tsx` (btw I think it should just be called `index.tsx` in cookie-cutter but there's already an `index.ts` in cdp-frontend).

hawkticehurst: I was also thinking about it more and I'm going to take back what I said about `create-react-app` being easiest. I think that will end being more boilerplate than we need and since we're already using `snowpack` here (which has the huge benefit of being a ridiculously fast dev server and very easy to set up).

The only thing snowpack doesn't do is provide bundling capabilities, so we would have to set up our own webpack or rollup bundling configuration (but I anticipate this wouldn't be too hard to do).

src/index.ts
hawkticehurst: Haha yeah I know I was very intrigued when I saw that pattern for the first time.

Also sounds good I'll just won't change anything and keep App as a named export for now then.

JacksonMaxfield:
Thanks for the changes @hawkticehurst. Merging. Will cut a release tomorrow / this weekend. And will likely have a PR on the cookiecutter repo for you and @tohuynh to take a look at to actually integrate this work with it.

index.html
hawkticehurst: Update: I was wrong about this last comment, snowpack can in fact bundle source code out of the box (they use `esbuild`). So I think continuing on with snowpack would be a very solid option.

PR #51, opened 2021-05-06
feature/home-search-bar

### LINKED ISSUE #37: Homepage Search Bar Component
Implement a homepage search bar component!

image

Please reference the UX Research on [homepage recommendations](https://docs.google.com/presentation/d/15rkic20QV6GU0_nL-8zHIgcw8-o1bCQIiYMsY2j6qFg/edit#slide=id.g8bf3178c59_0_43).

@JacksonMaxfield @tohuynh Any additional thoughts and comments for a future implementor?

tohuynh:
  Resolves #37 
  Add a search bar component with filters below. One filter allows users to search meetings, legislations, and/or council members. Another filter allows users to select a date range. (Though I'm not sure we can do date range filtering for legislations and council members?). Filtering components are assembled from `FilterPopup`, `SelectDateRange`, and `SelectTextFilterOptions`.
  On submitting the search form, it navigates the user to `/search` page, which will actually take care of querying the database. The search query and filters are stored in query parameters so that the `/search` page will know how to query the database. For example, a path to the search page could look like this: `/search?q=<search-query>&type=meeting,legislation&start=2020-02-02`.

Also, there were some minor changes to other files, just auto-formatting changes from `npm run format`.
  Just one story for [preview](https://tohuynh.github.io/cdp-frontend/?path=/story/library-layout-home-search-bar--default-story)

Thanks for contributing!


JacksonMaxfield (changes requested):
I mostly have some general design comments:

1. Can we add some padding around everything? The filters are so close together without any space to breath: 
image

2. Can we hide these filters under a "Advanced Options" dropdown? I kind of like the idea of hiding them by default as going back to the age old goal of "Search should be so good the general user doesn't need filters"

3. A part of me wants to say split up the text description / heading from this actual functional component. That's just design that could be placed on the larger App landing or something (idk thoughts @hawkticehurst).

4. In the search component example text, instead of "Keyword here", can we replace that with a random choice of like "sodo stadium" or "35 ave bike lane" or "capitol hill megablock" (insert a local issue here). I kind of like the idea of anyone who works on the front-end can add a local issue they care about to this list that gets chosen from to display the example text. Or a couple of examples: so the search bar would be: "35 ave bike line, capitol hill megablock, ..." The other option is to keep the current text as it is and just have these random options populate underneath the bar as small subtext as "Examples: ... "

5. Possible to make the Search Type changes as follows? 1. When there are options selected it should say: "Search Type(s): Meetings" instead of "Search Type(s): 1", and grow: "Search Type(s): Meetings, Legislation" and in the case that all are selected: "Search Type(s): All". And 2. We should never allow zero options to be selected so in the case none are selected can we have a little red text show up or something that says "Must select at least one search type".

annnnnddd finally, responding to your comments: Legislation filtering by date is def possible, if we use matter event linkage. Person is a bit more difficult.... But I think we have their start and end date separate from their created date (in Role) so it would be a joined query but possible for sure. Get all roles that are in the range, then find matching people to those roles.

Sorry this is a lot but just finally seeing _this_ component has me thinking a lot and I am very excited. Plus this like _the_ landing so its central to everything imo

src/components/Layout/HomeSearchBar/HomeSearchBar.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+const FilterContainer = styled.div({
+  display: "flex",
+  justifyContent: "center",
+  flexWrap: "wrap",
+});
+
+interface HomeSearchBarProps {
+  /**The municipal of the CDP instance.*/
+  municipal: string;
```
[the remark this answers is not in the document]
JacksonMaxfield: That is... make it `municipality` if we are keeping the pure design / text part of this component. I truly think that can go to some other container component or something. :shrug:

tohuynh:
>Can we hide these filters under a "Advanced Options" dropdown? I kind of like the idea of hiding them by default as going >back to the age old goal of "Search should be so good the general user doesn't need filters"

We could have something like Google, where `Tools` toggle the visibility of the filters? Any thoughts where Tools should be placed?
image

>A part of me wants to say split up the text description / heading from this actual functional component. That's just design that >could be placed on the larger App landing or something (idk thoughts @hawkticehurst).

That's fine too. Maybe that way we could reuse the search bar component for `/search` page.

>In the search component example text, instead of "Keyword here", can we replace that with a random choice of like "sodo >stadium" or "35 ave bike lane" or "capitol hill megablock" (insert a local issue here). I kind of like the idea of anyone who works >on the front-end can add a local issue they care about to this list that gets chosen from to display the example text. Or a >couple of examples: so the search bar would be: "35 ave bike line, capitol hill megablock, ..." The other option is to keep the >current text as it is and just have these random options populate underneath the bar as small subtext as "Examples: ... "

I'm not sure I follow. Do you mean to make the placeholder text (or Example options underneath the search bar) configurable as a prop?

>Possible to make the Search Type changes as follows? 1. When there are options selected it should say: "Search Type(s): >Meetings" instead of "Search Type(s): 1", and grow: "Search Type(s): Meetings, Legislation" and in the case that all are selected: >"Search Type(s): All". 

Sure, thankfully the only thing I need to add is a new render string function that is different from the general render string function.

>And 2. We should never allow zero options to be selected so in the case none are selected can we have a >little red text show >up or something that says "Must select at least one search type".

Good idea. Let's make all 3 search types selected as default or keep it at Meetings?

JacksonMaxfield:
>>In the search component example text, instead of "Keyword here", can we replace that with a random choice of like "sodo >>stadium" or "35 ave bike lane" or "capitol hill megablock" (insert a local issue here). I kind of like the idea of anyone who works >>on the front-end can add a local issue they care about to this list that gets chosen from to display the example text. Or a >>couple of examples: so the search bar would be: "35 ave bike line, capitol hill megablock, ..." The other option is to keep the >>current text as it is and just have these random options populate underneath the bar as small subtext as "Examples: ... "

> I'm not sure I follow. Do you mean to make the placeholder text (or Example options underneath the search bar) configurable as a prop?

Going to try ascii art:
```
[Search for a topic...                                   ]   [ Enter   ]
Example: 35th Ave Bike Lane                                  [ Filters ]
filters go here  *hidden by default
```

So the search bar still has placeholder "Search for a topic". But, underneath is like an italisized and `sub` tag or something with a random example of a topic one of the CDP devs cares about. We can ask everyone in slack about a local topic they care about :shrug: We store a list of example topics and each dev can give one or two and we just put a random choice there?



>>And 2. We should never allow zero options to be selected so in the case none are selected can we have a >little red text show >>up or something that says "Must select at least one search type".

> Good idea. Let's make all 3 search types selected as default or keep it at Meetings?

I say default to all. We will need to be good about async calls to DB then merge

JacksonMaxfield:
Actually update:

Going to try ascii art:
```
[Search for a topic...                                   ]   [ Enter   ]
Example: 35th Ave Bike Lane                                  [ Filters ]
```

If possible, it would actually be cool to replace the "Example" string with the filters instead of having them be below.
2 rows of stuff is better than 3 rows of stuff.

tohuynh:
> Actually update:
> 
> Going to try ascii art:
> 
> ```
> [Search for a topic...                                   ]   [ Enter   ]
> Example: 35th Ave Bike Lane OR filter components             [ Filters ]
> ```
> 
> If possible, it would actually be cool to replace the "Example" string with the filters instead of having them be below.
> 2 rows of stuff is better than 3 rows of stuff.

Oh hard code a couple of topics into the component and then randomly pick one to display as an example?

If filters are visible, the example string isn't visible?

JacksonMaxfield:
Yes just a hardcoded string that we randomly pick from. I just think it would be a fun addition :slightly_smiling_face: 

I think the Example should be below the search box because if the filters aren't showing then it would feel really imbalanced. And when the filters are showing, the example string is entirely gone.

tohuynh:
New commits:
- added some padding around some elements
- added an asset file with a list of example topics to be used as an example query
- removed the `Discover <municipality>...` heading
- added a different render string function. i.e. `Search Types(s): <selected-types>` instead of `Search Type: <num-selected>`
- added a conditional error msg `Please select at least one search type` in the popup itself
- the default selected search types are all 3.
---
New [preview](https://tohuynh.github.io/cdp-frontend/?path=/story/library-layout-home-search-bar--default-story)

It has two rows. One for the search bar(with small example query text underneath) and the filters row. The filters row has an `Advanced Options` button that toggles the filters' visibility. I decided not to toggle the example query text's visibility because the advanced options button isn't related to the example query text, and it would be puzzling to users as to why the example query text would appear/disappear.

JacksonMaxfield:
This looks so good! Last couple of comments:

Weird bug for me on Firefox Ubuntu (haven't checked other browsers)
image

When no meeting types are checked, and I hover the search bar, it fills with looking glasses :joy: 

Other than that, minor nitpick about adding a space after each comma of the list value. `Meetings, Legislations` instead of `Meetings,Legislations` (and on that note can we have it be just "Legislation" no "s" on the end).

Last and I guess this is just an option / question, did you try putting the advanced options on the same row as the example text but have the options grow left? Currently they are on a third row and grow right and just curious if you tried the other way. I just think having the Advanced options button directly underneath the search button would make the most visual sense.

JacksonMaxfield:
>> I don't think the Advanced Options button and Filter components should be on the same row as the example text because they look different in sizes and they aren't related to one another (one aguments searching while the other just provide helpful information).

> As a compromise, I can move the example text even closer (from 8px to 4px) to the search input and create space between the Advanced options button and the Filter components so that the advanced options button is underneath the search button:

That's a good reason for keeping them separate. In which case, just making the example text smaller but keeping the position and grow right behavior of the advanced options would probably be good.

So the only thing to change from the current commit would be to reduce example text size and add spaces after comma separated values in search types I believe.

src/components/Layout/HomeSearchBar/HomeSearchBar.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+const searchTypeOptions = [
+  {
+    name: SEARCH_TYPE.MEETING,
+    label: "Meetings",
+    disabled: false,
+  },
+  {
+    name: SEARCH_TYPE.LEGISLATION,
+    label: "Legislations",
```
JacksonMaxfield, line 56: Remove "s" from end.

src/components/Layout/HomeSearchBar/HomeSearchBar.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const selectedCheckboxes = Object.keys(checkboxes).filter((key) => checkboxes[key]);
+  let textRep = defaultText;
+  if (selectedCheckboxes.length === Object.keys(checkboxes).length) {
+    textRep += "(s): All";
+  } else if (selectedCheckboxes.length > 0) {
+    const selectedLabels = selectedCheckboxes.map((name) => {
+      const option = searchTypeOptions.find((option) => option.name === name);
+      return option?.label ?? "";
+    });
+    textRep += `(s): ${selectedLabels.join(",")}`;
```
JacksonMaxfield, line 82: `.join(", ")` (add a space)

JacksonMaxfield (approved):
Minor nitpicks but after earlier changes and discussion everything looks great!

tohuynh:
New commits:
- Decrease font size of exmaple text from 14px ish to 12px ish
- spaces after comma separated values in search types
- remove "s" from Legislations
- changes that reflects updated preview

Issue #55, opened 2021-06-03
Mobile-friendly Tables


BrianL3:
## Use Case
All tables in the project are currently bog-standard HTML, and make no reference to screen size.  If viewed on small screens, they become ugly-at-best, unusable-at-worst. 

## Solution
Probably turn the tables into cards of some sort. In reality, look up what the best practices are for mobile tables and... do that.

## Alternatives
I have no alternatives, go wild.

JacksonMaxfield:
Resolved in #63

PR #63, opened 2021-07-04
bugfix/mobile-friendly-tables

BrianL3:
[x] Name your pull request: Mobile Friendly Tables
[x] Link to any relevant issue in the PR description: [Mobile Friendly Tables](https://github.com/CouncilDataProject/cdp-frontend/issues/55) (Resolves #63 )
[x] Provide context of changes: Refactor of tables to accommodate mobile devices.
[x] Provide relevant tests for your feature or bug fix: see [here](https://brianl3.github.io/cdp-frontend/?path=/story/library-tables-meetingvotestable--single-broken-legislation-row)
[x] Provide or update documentation for any feature added by your pull request: Done


src/components/Tables/MeetingVotesTable/MeetingVotesTable.tsx
```diff
@@ -1,78 +1,54 @@
...
-        <th scope="col">Legislation</th>
-        <th scope="col">Council Decision</th>
-        <th scope="col">Votes</th>
-      </tr>
-    </thead>
-  );
-}
+const MeetingVotesTable = ({ votesPage }: MeetingVotesTableProps) => {
+  const COLUMN_NAMES = ["Legislation", "Council Decision", "Votes"];
+  const COLUMN_DISTRIBUTION = ["35%", "25%", "40%"];
```
BrianL3, line 19: The idea here is that Column distribution becomes a prop that will be passed down to lower components.  In the future, we can have user-adjustable column-size that leverages this. I wanted to lay the groundwork here but did not implement it as it would be separate feature.

src/components/Tables/ReactiveTableRow/ReactiveTableRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+          return (
+            <div
+              style={{
+                display: "flex",
+                justifyContent: "space-between",
+                flex: 1,
+                alignItems: "flex-start",
+              }}
+            >
+              <p
```
BrianL3, line 56: I did have a question I wanted to bring up: does anyone know something similar to `htmlFor` in React?  Basically I want to link the column name and content for accessibility purposes, but I am not sure how.

src/components/Tables/MeetingVotesTable/MeetingVotesTable.tsx
JacksonMaxfield: I love the idea of user adjustable column size!

JacksonMaxfield (approved):
From my non-JS-coding point of view / quick glance the code looks good to me!

Love the new design and layout. Love the idea / future planning on adjustable columns and such. Really great work.

tohuynh (approved):
Looks good overall to me.

Could you replace any mentions of 760px media queries with the tablet constant in styles/mediaBreakpoints.ts in all places this value is used?

Regarding your question linking column name and content, you could use dl, dt, dd HTML elements.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_htmlelementdiv_elements

You would be essentially treating the column name as the field name and content as the value. This way, you would treat the table row as a dl and a table cell as a pair of dt and dd. Then on desktop, you could make the dt element (field name or column name) have zero height and width, so that dt would be still available in the DOM for screen readers.

Issue #59, opened 2021-06-17
Add Event Container


JacksonMaxfield:
Create the container for a single event:

Example from v2:
https://councildataproject.github.io/seattle/#/events/81fae79b-02f5-4947-a0cf-99b30d51f42d
![image](https://user-images.githubusercontent.com/17132317/122328440-812ce300-cee4-11eb-9bd8-756218b9e6f2.png)

Updates for v3:
* Use new components
  * https://councildataproject.github.io/cdp-frontend/?path=/story/library-details-minutesitemslist--minutes-items-list
  * https://councildataproject.github.io/cdp-frontend/?path=/story/library-layout-tabbedcontainer--tabbed-container
  * https://councildataproject.github.io/cdp-frontend/?path=/story/library-tables-meetingvotestable--empty-legislation-vote-table
* Waiting on:
  * https://github.com/CouncilDataProject/cdp-frontend/issues/2

PR #72, opened 2021-08-02
component/event-video-player

### LINKED ISSUE #66: Meeting video player
## Use Case
Rewrite video player from seattle repo:
https://github.com/CouncilDataProject/seattle/blob/master/src/components/Event.jsx#L205
image

## Solution
For now, use [video-react](https://video-react.js.org/), which is currently used in seattle repo, and add a replay control as requested in [#75](https://github.com/CouncilDataProject/seattle/issues/75).

Add replay control documentation: https://video-react.js.org/components/replay-control/

Don't copy over the feature moving the video player to the right side of the page on scroll as most browsers have their own picture-in-picture functionality now.

## Alternatives

tohuynh:
  Resolves #66 

- [video-react](https://www.npmjs.com/package/video-react) is no longer being developed. Instead, we'll use [video.js](https://videojs.com/), which video-react was based on.
Unfortunately, video.js doesn't have a skip back/forward X seconds in its UI. For that I added a keyboard event handler to allow users to skip back/forward 10 seconds using arrow left and right, along with other standard keyboard events. `k` and `space` for play/pause. `m` for mute/unmute, and `f` for enter/exit full screen.

- I create a custom theme to style the video player, adapting from https://github.com/videojs/themes. I tried to make it like Youtube with the primary color being cdp_neon_blue.

- The player also supports language localization, by using `strings.getLanuage` and the player will use video.js language file.

- To create a 'jump to setence in video clip` call back for uss in #59 , I created a public api with a seekTo method.
```
const eventVideoRef = createRef<EventVideoRef>();
<EventVideo {..other props} componentRef={eventVideoRef} />
//create jump to video clip callback which then can then be passed to TranscriptFull and TranscriptSearch
const jumpToVideoClip = (startTime: number) => eventVideoRef.current?.seekTo(startTime)
```
https://tohuynh.github.io/cdp-frontend/?path=/story/library-details-event-video--default

Thanks for contributing!


hawkticehurst:
An immediate piece of feedback before a proper review is that the usage of `cdp_neon_blue` with the white foreground does not have enough contrast to be accessible, so I would pick a different background or foreground color.

tohuynh:
> An immediate piece of feedback before a proper review is that the usage of `cdp_neon_blue` with the white foreground does not have enough contrast to be accessible, so I would pick a different background or foreground color.

The video control component's background-color isn't entirely white, it's a linear gradient of black whose opacity gradually decreases as it moves toward the top of the component. I agree about the low contrast of white and cdp_neon_blue though in the big play button.

src/components/Details/EventVideo/EventVideo.tsx
```diff
@@ -0,0 +1,131 @@
...
+            timeDivider: true,
+            durationDisplay: true,
+            customControlSpacer: true,
+            remainingTimeDisplay: false,
+          },
+          preload: initialSeconds ? "auto" : "metadata",
+          aspectRatio: "16:9",
+          fluid: true,
+          language: strings.getLanguage(),
+          playbackRates: [0.75, 1, 1.5, 2],
```
JacksonMaxfield, line 82: Can we add 0.5 and reverse the order of this list. I am so used to the youtube version where it's the reverse order of this.

JacksonMaxfield (approved):
Only one change from me otherwise this is excellent!

src/components/Details/EventVideo/EventVideo.tsx
tohuynh: Yes to both

PR #78, opened 2021-08-22
feature/networking

BrianL3:
### Link to Relevant Issue

This pull request resolves the primary blocker for creating containers (e.g. most of the remaining work in the front end): a network layer. 

### Description of Changes

This is nearly purely additive. It adds a NetworkService which in turn will be used by other Services to request and create their appropriate Model objects. A "demo" of how this might work in included in PersonContainer for now. 

https://Brianl3.github.io/cdp-frontend/

### DIFF
--- a/src/components/PersonPage/PersonContainer.stories.mdx
+++ b/src/components/PersonPage/PersonContainer.stories.mdx
@@ -1,0 +1,31 @@
+import { action } from "@storybook/addon-actions";
+import { text, array, number, date } from "@storybook/addon-knobs";
+import { Meta, Props, Story } from "@storybook/addon-docs/blocks";
+
+import PersonContainer from "./PersonContainer";
+
+<Meta title="Library/Containers/Person" component={PersonContainer} />
+
+# PersonContainer
+
+### An Existing Person
+
+<Props of={PersonContainer} />
+
+<Story name="An Existing Person">
+  <PersonContainer
+    testPersonId="283ffb6c5a48"
+  />
+</Story>
+
+
+### A PersonContainer Error
+
+<Props of={PersonContainer} />
+
+<Story name="A Non-Existing Person">
+  <PersonContainer
+    testPersonId="bad-id"
+  />
+</Story>
+
--- a/src/components/PersonPage/PersonContainer.tsx
+++ b/src/components/PersonPage/PersonContainer.tsx
@@ -9,8 +9,8 @@ const PersonContainer: React.FC = () => {
   useEffect(() => {
     let personService = new PersonService();
     const fetchData = async () => {
-      let personId = window.location.pathname;
-      const result = await personService.getPerson(personId);
+      const slug = window.location.href.substring(window.location.href.lastIndexOf("/") + 1);
+      const result = await personService.getPerson(slug);
       setPerson(result);
     };
     fetchData();
--- a/src/components/PersonPage/PersonContainer.tsx
+++ b/src/components/PersonPage/PersonContainer.tsx
@@ -3,15 +3,34 @@ import Person from "../../models/Person";
 import { PersonService } from "../../networking/PersonService";
 import "@mozilla-protocol/core/protocol/css/protocol.css";
 
-const PersonContainer: React.FC = () => {
-  const [person, setPerson] = useState<Person | undefined>(undefined);
+type PersonContainerProps = {
+  /** optional person overrides fetch for testing */
+  testPerson?: Person;
+  testPersonId?: string;
+};
+
+const PersonContainer = ({ testPerson, testPersonId }: PersonContainerProps) => {
+  // 283ffb6c5a48
+  const [person, setPerson] = useState<Person | undefined>(testPerson);
+  const [error, setError] = useState<Error | undefined>(undefined);
 
   useEffect(() => {
     let personService = new PersonService();
     const fetchData = async () => {
       const slug = window.location.href.substring(window.location.href.lastIndexOf("/") + 1);
-      const result = await personService.getPerson(slug);
-      setPerson(result);
+      try {
+        if (testPersonId) {
+          const result = await personService.getPerson(testPersonId);
+          setPerson(result);
+          setError(undefined);
+        } else {
+          const result = await personService.getPerson(slug);
+          setPerson(result);
+          setError(undefined);
+        }
+      } catch (networkError) {
+        setError(networkError);
+      }
     };
     fetchData();
   }, []);
@@ -23,9 +42,14 @@ const PersonContainer: React.FC = () => {
           {`${JSON.stringify(person)}`}
         </p>
       )}
-      {!person && (
+      {!person && !error && (
         <p className="mzp-c-card-desc" style={{ marginLeft: 8 }}>
           No response yet
+        </p>
+      )}
+      {error && (
+        <p className="mzp-c-card-desc" style={{ marginLeft: 8 }}>
+          {error.message}
         </p>
       )}
     </div>
--- a/src/components/PersonPage/PersonContainer.tsx
+++ b/src/components/PersonPage/PersonContainer.tsx
@@ -15,7 +15,7 @@ const PersonContainer = ({ testPerson, testPersonId }: PersonContainerProps) => 
   const [error, setError] = useState<Error | undefined>(undefined);
 
   useEffect(() => {
-    let personService = new PersonService();
+    const personService = new PersonService();
     const fetchData = async () => {
       const slug = window.location.href.substring(window.location.href.lastIndexOf("/") + 1);
       try {
--- a/src/components/PersonPage/PersonContainer.tsx
+++ b/src/components/PersonPage/PersonContainer.tsx
@@ -1,15 +1,15 @@
 import React, { useEffect, useState } from "react";
 import Person from "../../models/Person";
 import { PersonService } from "../../networking/PersonService";
+import { useLocation } from "react-router-dom";
 import "@mozilla-protocol/core/protocol/css/protocol.css";
 
 type PersonContainerProps = {
   /** optional person overrides fetch for testing */
   testPerson?: Person;
-  testPersonId?: string;
 };
 
-const PersonContainer = ({ testPerson, testPersonId }: PersonContainerProps) => {
+const PersonContainer = ({ testPerson }: PersonContainerProps) => {
   // 283ffb6c5a48
   const [person, setPerson] = useState<Person | undefined>(testPerson);
   const [error, setError] = useState<Error | undefined>(undefined);
@@ -17,17 +17,11 @@ const PersonContainer = ({ testPerson, testPersonId }: PersonContainerProps) => 
   useEffect(() => {
     const personService = new PersonService();
     const fetchData = async () => {
-      const slug = window.location.href.substring(window.location.href.lastIndexOf("/") + 1);
+      const slug = useLocation().pathname.replace("/", "");
       try {
-        if (testPersonId) {
-          const result = await personService.getPerson(testPersonId);
-          setPerson(result);
-          setError(undefined);
-        } else {
-          const result = await personService.getPerson(slug);
-          setPerson(result);
-          setError(undefined);
-        }
+        const result = await personService.getPerson(slug);
+        setPerson(result);
+        setError(undefined);
       } catch (networkError) {
         setError(networkError);
       }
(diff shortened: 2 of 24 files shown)


src/components/PersonPage/PersonContainer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const PersonContainer = ({ testPerson, testPersonId }: PersonContainerProps) => {
+  // 283ffb6c5a48
+  const [person, setPerson] = useState<Person | undefined>(testPerson);
+  const [error, setError] = useState<Error | undefined>(undefined);
+
+  useEffect(() => {
+    let personService = new PersonService();
+    const fetchData = async () => {
+      const slug = window.location.href.substring(window.location.href.lastIndexOf("/") + 1);
```
JacksonMaxfield, line 20: We had a request a long time ago to have both the id and router string be available. i.e. both some sha value and `/lorena-gonzalez` for example. That value _should_ always be available for a person and it _should_ always be unique. Is there a way to check for the "router_string" in addition to the id?

Also, I know this PR is just focused on the networking layer but just curious.

src/models/Body.ts
```diff
@@ -0,0 +1,32 @@
+import { ResponseData } from "../networking/NetworkResponse";
```
JacksonMaxfield, line 1: Going to skip over all of these for now, they all look fine, it would be good to eventually move these to be auto-generated files.

Maybe a project would be to split out the database models from cdp-backend into it's own repo as JSON files and have `cdp-backend` and `cdp-frontend` auto-generate these types of model classes.

Again this is way way down the line.

src/components/PersonPage/PersonContainer.stories.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+### A PersonContainer Error
+
+<Props of={PersonContainer} />
+
+<Story name="A Non-Existing Person">
+  <PersonContainer
+    testPersonId="bad-id"
+  />
+</Story>
```
JacksonMaxfield, line 30: Do we know how storybook reacts to a database request?

JacksonMaxfield: Any way to mock this behavior?

JacksonMaxfield: Or... I guess once we have the integration repo up-and-running again we can just point these requests to the integration database.

src/networking/NetworkService.ts
```diff
@@ -0,0 +1,123 @@
...
+  private static instance: NetworkService;
+  private static db: FirebaseFirestore;
+  /**
+   * The NetworkService's constructor should always be private to prevent direct
+   * construction calls with the `new` operator.
+   */
+  private constructor() {
+    // initialize firebase
+    const firebaseApp = initializeApp({
+      projectId: "cdp-jackson-dev-002",
```
JacksonMaxfield, line 32: Can this be parametrizable somehow? How should we integrate this with the cookiecutter? A JSON config somewhere or just a value somewhere?

src/networking/PersonService.ts
```diff
@@ -0,0 +1,27 @@
+import { NetworkService } from "./NetworkService";
+import { COLLECTION_NAME } from "./PopulationOptions";
+import Person from "../models/Person";
+
+export class PersonService {
+  networkService: NetworkService = NetworkService.getInstance();
```
JacksonMaxfield, line 6: So if I understood your comments on the base service, this call will either use the cached instance or make a new one correct?

JacksonMaxfield (commented):
This all looks good to me and makes sense. Some random questions.

Before approve, any chance we can fix the lint and build?

src/networking/NetworkService.ts
BrianL3: Just from what you have said, it should probably write a JSON file with this (and other) values for us to pick up. That sounds easy enough to utilize on our end. Would that be a problem?

src/networking/PersonService.ts
BrianL3: Yes. Since every model will have its own Model Service (e.g. PersonService) and every Model Service has a NetworkService... just have them all use the same one, it's cheaper that way.

src/networking/NetworkService.ts
JacksonMaxfield: Nope totally fine!

JacksonMaxfield (approved):
Looks good to me. @tohuynh should get a change to review though.

src/components/PersonPage/PersonContainer.stories.mdx
BrianL3: Storybook works with the DB request. It pulls and prints just fine. 

I do want to make mocks though, for the Services. I don't know how to do it cleanly though - something like I did seems very clumsy. I am open to suggestions.

src/models/Body.ts
BrianL3: That would indeed be nice

src/components/PersonPage/PersonContainer.stories.mdx
JacksonMaxfield: The "easy way" to do it is just have another parameter at the end with `mockData` that is an object itself. Idk how to do mocks otherwise...

JacksonMaxfield: for storybook anyway

JacksonMaxfield: Maybe @hawkticehurst knows?

src/components/PersonPage/PersonContainer.tsx
BrianL3: That sounds like it would be very difficult to distinguish a `router_string` vs the ID of the person object, if they are both strings and live in the same place (the slug).

JacksonMaxfield: Check a substring of the slug?

`id--{some_id}` vs `route--{some_router_string}` maybe?

src/components/PersonPage/PersonContainer.stories.mdx
BrianL3: I have done the "easy" way before - the Model Service just checks for mockData and does an early return with whatever object passed to it.  But that doesn't cleanly mesh with the Storybook.  

Perhaps the following?
Containers have optional props for the things that they fetch - e.g. a Person? for a PersonContainer. That is their mock object for Storybook purposes.  And then Services test in a different way.  IIRC it is possible to have an entire mock Firestore instance that spins up locally, that would probably be ideal. Then we don't have to worry about our testing flow avoiding the actual nuts and bolts of the function.

src/components/PersonPage/PersonContainer.tsx
BrianL3: Wouldn't that require that all IDs and all `router_strings` have prefixes appended to them?

JacksonMaxfield: Why not in this slug parsing code check for those prefixes?

If it's using the `id--` prefix then do exactly what you are doing now but strip the prefix first.

If it's using the `route--` prefix then query for docs with that `router_string`, if more than one or none, error. if just one then use the data received?

src/components/PersonPage/PersonContainer.stories.mdx
JacksonMaxfield: Yea I guess using the firestore mock instance is a better solution.. How easy is it I guess?

JacksonMaxfield: But yes. for now, I say that the mock data can live on the service level.

src/components/PersonPage/PersonContainer.tsx
tohuynh: Since we are using react-router-dom, we should eventually use react-router-dom `useLocation` to get the slug from its [location](https://reactrouter.com/web/api/location). But this isn't particularly important for this PR.

Regarding prefix, we could switch to `/person?id=<person-id>` or `/person?router_string=<router-string>`, instead of `/person/<slug>`

src/models/Body.ts
tohuynh: Agree with Jackson, to avoid manually keeping these files and database models in sync.

src/components/PersonPage/PersonContainer.stories.mdx
```diff
@@ -0,0 +1,31 @@
+import { action } from "@storybook/addon-actions";
+import { text, array, number, date } from "@storybook/addon-knobs";
+import { Meta, Props, Story } from "@storybook/addon-docs/blocks";
```
tohuynh, line 3: Just FYI addon-knobs will be deprecated soon and we should stop using it to not create any more work for #21.

Also, I stopped writing stories in .mdx and started using .tsx (for example, this [story](https://github.com/CouncilDataProject/cdp-frontend/blob/main/src/components/Details/EventVideo/EventVideo.stories.tsx)). My editor doesn't provide any auto-suggestion for mdx files. You can still write .mdx file though.

src/components/PersonPage/PersonContainer.stories.mdx
tohuynh: Agree, mock firestore instance would be ideal. But your solution is OK for now.

src/networking/NetworkService.ts
```diff
@@ -0,0 +1,123 @@
...
+    documentId: string,
+    collectionName: COLLECTION_NAME,
+    populationOptions?: PopulationOptions
+  ): Promise<NetworkResponse> {
+    const documentRef = doc(NetworkService.db, collectionName, documentId);
+    const response = new NetworkResponse();
+    return getDoc(documentRef)
+      .then((docSnap) => {
+        if (!docSnap || !docSnap.data()) {
+          return Promise.reject(
```
tohuynh, line 94: Better to use !docSnap.exists() here as the condition?

src/networking/NetworkService.ts
```diff
@@ -0,0 +1,123 @@
...
+            new Error(`Document ${documentId} does not exist in Collection ${collectionName}.`)
+          );
+        }
+        response.data = docSnap.data();
+        if (populationOptions && populationOptions.toPopulate) {
+          const cascade: Promise<NetworkResponse>[] = [];
+          const refsToPopulate: string[] = [];
+          populationOptions.toPopulate.forEach((docRefToPopulate: Populate) => {
+            const refValueToPopulate = response.data![docRefToPopulate.refName];
```
tohuynh, line 103: At this point response.data is non-null right?
Do we need the `!.`?

tohuynh (approved):
The NetworkService and PersonService look good to me.

Just have a few comments.

src/components/PersonPage/PersonContainer.tsx
tohuynh: You could parse the slug to determine the prefix too.

src/networking/NetworkService.ts
BrianL3: That's what I thought, but the compiler was complaining.  Looking at it again, I realize what is wrong.

src/networking/NetworkService.ts
BrianL3: yep, done

src/models/Body.ts
BrianL3: @JacksonMaxfield Can you make an issue for this - it doesn't really belong on cdp-frontend's board solo.  I'm happy to do the work once we have either a simple flatfile or separate repo. 
My plan would be to have a node script that writes new model files based on the schema provided by the other repo or flatfile.

src/components/PersonPage/PersonContainer.tsx
BrianL3: I switched to useLocation for now.  We'll refactor as necessary.

src/components/PersonPage/PersonContainer.stories.mdx
BrianL3: I'm looking into [a potential solution](https://www.npmjs.com/package/mock-firebase-ts) for a mock firebase.  Perhaps we can set it up to run as part of storybook's run script, and have NetworkService swap to using the mock on some environmental variable?

src/components/PersonPage/PersonContainer.stories.mdx
BrianL3: I'll start using .tsx instead, thanks for lookin' out.

src/models/Body.ts
JacksonMaxfield: Yep. I will add one tonight!

JacksonMaxfield:
@BrianL3 are you going to try and get the mock firebase into this PR or is this good to go for now?

src/models/Body.ts
JacksonMaxfield: Created and tracking here: https://github.com/CouncilDataProject/cdp-roadmap/issues/13

PR #85, opened 2021-09-11
feature/event-container

### LINKED ISSUE #59: Add Event Container

tohuynh:
### Link to Relevant Issue

This pull request resolves #59 

### Description of Changes

PR has 4 main parts.

#### Event Page and Container
Created a folder for pages and containers. `EventPage` is where the data is fetched by using all the XServices and given to `EventContainer` to assemble and render all the relevant components.

#### General Fetch Data Container
Created a general fetch data container to handle UI changes from data fetching. The UI needs to reflect the current state (loading, has an error, or has data) of data fetching. This section comprises of two parts - a component `FetchDataContainer` and a hook `useFetchData`. Example use of hook
```
const { state: eventDataState, dispatch: eventDataDispatch } = useFetchData<EventData>({
    isLoading: false,
  });
````
The `eventDatastate` represents the current state of fetching event data and the `eventDataDispatch` provides a way to update the state. For examples of using the dispatch see the `EventPage`. The `FetchDataContainer` just renders different things depending on the `eventDatastate`. Show loader when it is loading, show error for has an error, and show `EventContainer` for has data.

#### Download transcript json service
Created a `TranscriptJson` model is that is similar to the other database models. Also moved transcript types (Sentence and Word) into this file. To get a TranscriptJson, I added a `downloadJson` method that returns a `NetworkResponse` and a `TranscriptJsonService` that calls this method to download the JSON content. The `TranscriptJsonService` method behaves like other XService methods.

#### Multiple Sessions UI
For the multiple session videos, I place them within a tab component. The tab menu titles are the session_datetimes. Changing the current session tab will pause the previous video player. Clicking on `jump to X` will also pause the previous video player if the user is jumping to a new session.

Do we need the transcripts to reflect the multiple sessions? Meaning should the full transcript be separated into different sessions (like the videos are) or just combined into one list of transcript items? Maybe if we change the hh:mm:ss(start_time) on these transcript items to the actual time using the session_datetime that would provide session context for the transcript items? This assumes that session_datetime and video start time are the same, is this true?

https://tohuynh.github.io/cdp-frontend/?path=/story/library-containers-event--default

### CODE REVIEW

File: src/networking/VoteService.ts [hunk 1]
```diff
@@ -26,6 +26,7 @@ export default class VoteService extends ModelService {
           WHERE_OPERATOR.eq,
           doc(NetworkService.getDb(), COLLECTION_NAME.Event, eventId)
         ),
+        orderBy(REF_PROPERTY_NAME.VotePersonRef),
```
JacksonMaxfield, line 29: I assume this is to get the same ordering every time the page loads?

src/containers/FetchDataContainer/FetchDataContainer.tsx
```diff
@@ -0,0 +1,27 @@
...
+  children,
+  error,
+}: FetchDataContainerProps) => {
+  if (isLoading) {
+    return <Loader />;
+  }
+  if (error) {
+    // Display the error for now.
+    // TODO: throw the error and catch it with a general error boundary that displays the error page.
+    return <span>{error}</span>;
```
JacksonMaxfield, line 22: It would be nice to have a task for after v3 of making this a pretty error or something... Just like providing something for the user to look at besides a span.

JacksonMaxfield (commented):
First off. Omg this looks great. All the new designs together is wonderful.

Early review:
* nothing sticks out about the code, it all looks fine
* minor nitpicks on components:
  * can you add like 4px of right padding to the sentence components, when i scroll on either the transcript search of the full transcript, the jump to icon is covered by the scroll bar.
  * can you include the session index in the session tabs? `"Session 1: 4:00pm"` it's currently not clear to the user what those mean and in the case that session time is the same between sessions it will help a bit... I might argue to get rid of displaying the time all-together and just showing `"Session 1"` `"Session 2"` even because of no data validation on the session time.
  
Answering your questions about how to handle transcripts for sessions. First, yes, there are multiple transcripts for an event. One transcript per session. Unlike the video however, I do like the idea of combining the transcripts somehow. Makes it easier to search and such. I would say, just like how we used to have the "New Speaker" marks on the transcript UI, we could just have a "Session 1" and "Session N" horizontal break to indicate it.

src/containers/FetchDataContainer/FetchDataContainer.tsx
BrianL3: I expect we'll make a general Error component that pops a modal or something.

src/components/Shared/ResponsiveTab.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Tab } from "semantic-ui-react";
+import { fontSizes } from "../../styles/fonts";
+import { screenWidths } from "../../styles/mediaBreakpoints";
+
+interface ResponsiveTabProps {
+  padding: string;
+  breakpoint?: string;
+}
+
+export default styled(Tab)<ResponsiveTabProps>((props) => ({
```
BrianL3, line 12: I appreciate the detailed comments here.

src/utils/createError.ts
```diff
@@ -0,0 +1,11 @@
+/** Create Error from error of unknown type */
+
+export function createError(error: unknown): Error {
```
BrianL3, line 3: This is basically just a Model class, no? I'm fine with it being a `util` thing as well. I think in certain edge cases you get back full HTML pages from broken requests as well, we might want to test this against that scenario.

BrianL3 (approved):
Took me a while to grok the architecture here but I think I got it:
EventPage creates a FetchDataContainer, which fetches the Event.
FetchDataContainer handles error stuff, if any.
The child component (EventContainer) is guaranteed the Event data by prop.  

I like it a lot; great work.

src/networking/VoteService.ts
tohuynh: Yes, I think so.

src/utils/createError.ts
tohuynh: The `createError` util has only one use case. In `try catch` blocks, the caught error is of unknown type and my editor was complaining I can't set the error.message on an unknown type, so I had to convert to an `Error`. I'll add a comment to the util description.

tohuynh:
> EventPage creates a FetchDataContainer, which fetches the Event.
> FetchDataContainer handles error stuff, if any.
> The child component (EventContainer) is guaranteed the Event data by prop.

The FetchDataContainer is even simpler than that. It basically does conditional rendering depending on the value of `eventDataState`. Render the loader if data is currently being fetched. Renders an error message if there's an error. Or render the EventContainer(provided as a child component) if data was successfully fetched.

tohuynh:
>can you add like 4px of right padding to the sentence components, when i scroll on either the transcript search of the full transcript, the jump to icon is covered by the scroll bar.

Yes

>can you include the session index in the session tabs? "Session 1: 4:00pm" it's currently not clear to the user what those mean and in the case that session time is the same between sessions it will help a bit... I might argue to get rid of displaying the time all-together and just showing "Session 1" "Session 2" even because of no data validation on the session time.
Answering your questions about how to handle transcripts for sessions. First, yes, there are multiple transcripts for an event. One transcript per session. Unlike the video however, I do like the idea of combining the transcripts somehow. Makes it easier to search and such. I would say, just like how we used to have the "New Speaker" marks on the transcript UI, we could just have a "Session 1" and "Session N" horizontal break to indicate it.

I like the `Session N` in the tab menu title. I want to keep the tab menu title short so we can display two tab menu titles horizontally on mobile. I'll also add the hours to the event_datetime string.

> Makes it easier to search and such. I would say, just like how we used to have the "New Speaker" marks on the transcript UI, we could just have a "Session 1" and "Session N" horizontal break to indicate it.

I could add `Session N` to the transcript item itself, in between the speaker name and the hh:mm:ss. I think this way is better than horizontal breaks because it reduces cognitive load. When the horizontal breaks are out of view, users wouldn't have to remember which session they are in as they read the transcript.

JacksonMaxfield:
> I like the `Session N` in the tab menu title. I want to keep the tab menu title short so we can display two tab menu titles horizontally on mobile.

Makes sense!



> I could add `Session N` to the transcript item itself, in between the speaker name and the hh:mm:ss. I think this way is better than horizontal breaks because it reduces cognitive load. When the horizontal breaks are out of view, users wouldn't have to remember which session they are in as they read the transcript.

Love it.

Issue #60, opened 2021-06-17
Add Person Container


JacksonMaxfield:
Add the person container

example from v2:
https://councildataproject.github.io/seattle/#/people/d7d09ac5-b322-435c-b709-9851789d0323
![image](https://user-images.githubusercontent.com/17132317/122328890-4bd4c500-cee5-11eb-9b4a-e71b606da380.png)

Here was a mockup from design team for inspiration:
![image](https://user-images.githubusercontent.com/17132317/122328873-424b5d00-cee5-11eb-82fa-1522b5c44639.png)

The design team is close to perfect, although we won't be doing anything with twitter or instagram feeds.
Needs for v3 upgrade:
* a design for the council member information (we have WAY more of it in the v3 database) see the schema: https://councildataproject.github.io/cdp-backend/database_schema.html
  * we can list the person roles, their tenure, their chairing, etc.
* use the new voting record table: https://councildataproject.github.io/cdp-frontend/?path=/story/library-tables-votingrecordtable--many-votes

We are waiting on:
* https://github.com/CouncilDataProject/cdp-frontend/issues/55

Note: the map / district image is optional in the database so designs should be able to have that or not.

JacksonMaxfield:
Also the voting records table should be filterable and paginated: https://github.com/CouncilDataProject/cdp-frontend/issues/54

JacksonMaxfield:
@tohuynh had a recommendation on the prior version of this issue:

> Just a thought, some mention of the voting records table should be initially visible (even if clicking on it just scrolls down to the table).

PR #91, opened 2021-09-16
feature/firebase-config

### LINKED ISSUE #86: Add firebase config to exportable app
### Feature Description

Allow the cookie cutter to inject firebase config to the exportable app.
Once the app has access to the firebase config, inject the config into the network layer to connect the network layer to firebase/firestore.

### Use Case

Different deployments have different firebase config. The app should take the config as a prop.

### Solution

### Alternatives

_Please describe any alternatives you've considered, even if you've dismissed them._

### Screenshots/References

_In the case of a visually oriented feature or component request, providing drawings, screenshots, and/or links to examples of what you're seeking will be very helpful._

tohuynh:
### Link to Relevant Issue

This pull request resolves #86 

### Description of Changes

Add a React context(a way to pass data through the component tree without manually passing down props) to contain the firebase config (and possibly other configurations in the future). For us, they are global static data such as firebase config, information about the deployment such as the municipal's name (for use on the home page), and maybe even styling themes.

Then any component that subscribes to this React context, can use a hook `useAppConfigContext` to use the context values. A component can subscribe to the React context by being a descendent of the `AppConfigProvider` component.

In cookie cutter it will look something like this:
`<AppConfigProvider appConfig={ actual app config}><App /><AppConfigProvider />`

In the PR, I added `useAppConfigContext` to get the firebase configs in the `EventPage` and `PersonPage` and pass them to the network layer.

I also copied over the data fetching architecture from the `EventPage` to the `PersonContainer` and `PersonPage`. However, the data fetched in the `PersonPage` is currently insufficient to render the person container. That's saved for #60.


src/pages/PersonPage/PersonPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+        if (!didCancel) {
+          personDataDispatch({
+            type: FetchDataActionType.FETCH_SUCCESS,
+            payload: person,
+          });
+        }
+      } catch (err) {
+        if (!didCancel) {
+          const error = createError(err);
+          personDataDispatch({ type: FetchDataActionType.FETCH_FAILTURE, payload: error });
```
JacksonMaxfield, line 48: I think I just found a typo:

`FETCH_FAILTURE` -> `FETCH_FAILURE`

JacksonMaxfield (approved):
This literally all looks perfect besides for the typo I found.

Amazing.

BrianL3 (approved):
Thanks To.  I was midway through the PersonContainer stuff with the new pattern but this is fine, speeds things up really. 

Passing the config through each Page is not ideal IMO, but I can see why its set up this way and have no better idea.  Optimally, I would think, only the NetworkService class would actually have any idea about the Firestore context to keep it completely decoupled from the view layers.  But it seems difficult to get it there.  Can useAppConfigContext not be imported into NetworkServices directly (as they are not React components?)?

tohuynh:
> Can useAppConfigContext not be imported into NetworkServices directly (as they are not React components?)?
Yeah, useAppConfigContext is React hook and can be used only in React functional components.

PR #96, opened 2021-09-22
bugfix/handle-no-doc-return-and-correct-2d-array-init

JacksonMaxfield:
### Description of Changes

Found during testing a new dev infra after completing: https://github.com/CouncilDataProject/cdp-backend/pull/111

When I tried loading an event page it would fail because of these two bugs.


JacksonMaxfield:
Going to merge so that I can cut a new release to work on the cookiecutter. If this is a bug or shouldn't be committed we can revert later.

PR #106, opened 2021-10-14
feat/change document.title on the person and event pages

### LINKED ISSUE #93: Add page title to pages
Add pages to the different pages.

Use this react hook: https://github.com/CouncilDataProject/seattle/blob/master/src/hooks/useDocumentTitle.js

zmbc:
### Link to Relevant Issue

This pull request resolves #93.

### Description of Changes

Adds a hook to change the page title, and uses it from the Person and Event pages (the only two pages currently implemented that need a title, I think).

Not a super visual change, but you can see it in action by clicking the little "pop-out" button in the top right of Storybook.

[Storybook for Event](http://zebburkeconte.com/cdp-frontend/index.html?path=/story/library-containers-event--event)

[Storybook for Person](http://zebburkeconte.com/cdp-frontend/index.html?path=/story/library-containers-person--default)

### DIFF
--- a/src/pages/HomePage/HomePage.tsx
+++ b/src/pages/HomePage/HomePage.tsx
@@ -3,6 +3,7 @@ import styled from "@emotion/styled";
 
 import { HomeSearchBar } from "../../components/Layout/HomeSearchBar";
 
+import useDocumentTitle from "../../hooks/useDocumentTitle";
 import { strings } from "../../assets/LocalizedStrings";
 
 const Container = styled.div({
@@ -11,6 +12,8 @@ const Container = styled.div({
 });
 
 const HomePage: FC = () => {
+  useDocumentTitle(strings.council_data_project);
+
   return (
     <Container>
       <h1 className="mzp-u-title-xs">{strings.search_city_council}</h1>


zmbc:
Thanks for the reviews! It actually wasn't possible to do both things exactly the way you requested them, but I tried to find a compromise.

Also, if the document title was supposed to be localized, that should be done at the whole-app level (currently it uses a non-localized string directly in the HTML) so I added that.

tohuynh:
A useEffect hook has two parts: the effect and a clean-up of the effect.

In our case, the effect is changing the document.title depending on the value of the `title` arg. The clean-up happens when the component unmounts, and the document.title changes to a default title, that would "Council Data Project".

useDocumentTitle looks a little off with the initial empty function return -- since returning a function is reserved for the clean-up function. 
```
import { useEffect } from "react";

function useDocumentTitle(title?: string) {
  useEffect(() => {
    if (title) {
      // only run the effect if the title arg is truthy (non-null, non-empty, not undefined)
      document.title = title;
    }

    return () => {
      // clean up the effect
      document.title = strings.council_data_project;
    };
  }, [title]);
}

export default useDocumentTitle;
```

zmbc:
@tohuynh I understood that. I think I'm looking at the problem a bit differently.

The way I see it, the point of `useDocumentTitle` is to set the document title to something *while a component is mounted*. So, my implementation sets the title, then re-sets it to whatever it was previously as cleanup.

When the function is called with a blank or null title, it does nothing. Does not change the title, and does not have to clean up. That is why there is an early return. I have just made a slight change to return without a cleanup function however, as it is not needed to supply a no-op function.

tohuynh:
>So, my implementation sets the title, then re-sets it to whatever it was previously as cleanup.

OK that makes sense. The document.title is quarnateed to go back to "CDP" when `title` is falsey.

Could you use strict equality (===) for the null check? Since you made `title` optional, check for `title === undefined` as well.

zmbc:
> Could you use strict equality (===) for the null check? Since you made title optional, check for title === undefined as well.

[== null is a shorthand for both.](https://stackoverflow.com/a/5101991/5981641)

> I think it's better to couple with a constant value than having some of useDocumentTitle calls rely on another useDocumentTitle call in App.tsx to produce the correct document.title.

There will have to be nested useDocumentTitle calls anyway, because that is the only way to localize the document title (which is currently not localized).

If it's coupled, if the "default" document title ever changes, we will have to change it in two places. If we forget, it will cause very confusing bugs.

The "setting it back" way is more conceptually clear because that's what "use document title" means: use this document title until I'm not rendered anymore. Not: use this document title until I'm not rendered anymore, at which point set the document title to this other value. It also supports useDocumentTitle by arbitrarily deeply nested components, which the `strings.council_data_project` approach doesn't (useDocumentTitle by siblings is not supported by either approach).

zmbc:
> Could you add a useDocumentTitle to the Homepage?

Done.

PR #146, opened 2021-12-17
fix/minutes-item-expansion

### LINKED ISSUE #126: Minutes Item expansion cleanup
1. Remove the minutes item arrow right: 
image

We have a plus sign, no need for the arrow as well.

2. Why is the "See documents" split across two lines?

3. If the minutes item has a description, we should display description before all of the attachments. I hacked this into the html of the page but here is a rough example:

image

4. All links should be made to open in tab instead of redirecting from the current CDP page.

tohuynh:
### Link to Relevant Issue

This pull request resolves #126 

### Description of Changes

- Bold the minutes item's name to distinguish it from minutes item's description
- Add the minutes item's description
- "See documents" color is dark blue
- Remove the triangle marker (arrow right) from "See documents"
- Place the +/- icon immediately after "see documents" and make it all one line.
- Place the attachment links in `<ul>` to add indentation from the minutes item's name and description
- The attachment URL open in a new tab

### Screenshot

image

### DIFF
--- a/src/components/Details/MinutesItemsList/DocumentsList.tsx
+++ b/src/components/Details/MinutesItemsList/DocumentsList.tsx
@@ -12,11 +12,8 @@ const Summary = styled.summary({
   color: colors.dark_blue,
   display: "flex",
   alignItems: "center",
-  "&::before": {
+  "&::before, &::marker": {
     // remove mozilla protocol +/- icon
-    content: "none",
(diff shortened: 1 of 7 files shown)


JacksonMaxfield (approved):
This looks _great_! Thanks! Merge whenever!

PR #148, opened 2021-12-28
feature/lazy-load-transcript-in-event-page

### LINKED ISSUE #101: Allow event page to render before transcript arrives
### Feature Description

_A clear and concise description of the feature you're requesting._

Allow the event page / container to render with basic info before the full transcript JSON arrives back from request to GCS

### Use Case

_Please provide a use case to help us understand your request in context._

Reduce the pages "time to interactivity". Looking at the network monitor, everything besides the transcript returns in ~300ms which is _okay_ and we should look into how to speed this up further but alright for now.

### Solution

Only load things as they come in. I.e. Maybe the transcript search and transcript tabs show a "Fetching transcript..." line until they have returned?

tohuynh:
### Link to Relevant Issue

This pull request resolves #101 

### Description of Changes

The event page can load before the transcript JSON is retrieved. While the transcript is loading "Fetching transcript..." is shown in Search transcript component (where `N result(s)` is shown, to the right of `Search transcript`) and Full transcript component.

- Separate each call to the network layer(to get all the data for the event page) into its own `useFetchData`. The call to fetch transcript JSON runs when the call to fetch `sessions` is finished.
- Added some optimization with `useMemo` in `EventInfoTabs`
- Combined `eventMinutesItems` and `eventMinutesItemsFiles` into one prop.

### DIFF
--- a/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
+++ b/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
@@ -55,7 +55,7 @@ manySentences.args = {
     session_index: 0,
     index: i,
     start_time: i,
-    text: `This is a sentence${i}.`,
+    text: `This is a sentence ${i}.`,
     speaker_name: "Lisa Herbold",
     speaker_id: "lisa-herbold",
     speaker_index: 0,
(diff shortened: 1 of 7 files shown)


JacksonMaxfield (approved):
This all looks good to me! Merge whenever.

JacksonMaxfield:
Also this is a big optimization so thanks for handling it! ❤️

PR #149, opened 2021-12-28
feature/share-event-page

### LINKED ISSUE #95: Add share at timepoint button to event page
Add a share button with "start at timepoint" functionality to the event page.

tohuynh:
### Link to Relevant Issue

This pull request resolves #95 

### Description of Changes

Add a share button to generate a shareable URL for an event, with optional query parameters for the current session and user-selected timepoint.

This is mostly a copy of the original [functionality](https://github.com/CouncilDataProject/seattle/blob/master/src/components/EventShareVideo.jsx) with some changes to make it type-safe and add the current session to the shareable URL.

### Screenshots
(share button)
image

(modal)
image

### DIFF
--- a/src/pages/EventPage/EventPage.tsx
+++ b/src/pages/EventPage/EventPage.tsx
@@ -50,7 +50,8 @@ const EventPage: FC = () => {
     return [s, t].map((el) => {
       if (el !== null && typeof el === "string") {
         const n = parseFloat(el);
-        return isNaN(n) ? 0 : n;
+        const isInvalid = isNaN(n) || n < 0;
+        return isInvalid ? 0 : Math.floor(n);
       } else {
         return 0;
       }
--- a/src/pages/EventPage/EventPage.tsx
+++ b/src/pages/EventPage/EventPage.tsx
@@ -1,4 +1,5 @@
 import React, { FC, useCallback, useMemo, useEffect } from "react";
+import queryString from "query-string";
 import { useParams, useLocation } from "react-router-dom";
 
 import Event from "../../models/Event";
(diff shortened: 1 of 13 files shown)


src/pages/EventPage/EventPage.tsx
```diff
@@ -44,6 +45,18 @@ const EventPage: FC = () => {
...
 
+  const [initialSession, initialSeconds] = useMemo(() => {
+    const { s, t } = queryString.parse(location.search);
+    return [s, t].map((el) => {
+      if (el !== null && typeof el === "string") {
+        const n = parseFloat(el);
+        return isNaN(n) ? 0 : n;
+      } else {
+        return 0;
+      }
```
JacksonMaxfield, line 56: My only question for the whole pull request is "what happens if someone shares `url?s=1` will that route to session 1 and default to 0 seconds? AND vice-versa, if someone only shares the seconds `url?t=100` will that default to session=0?

I think this is the code block that does that but just want to make sure.

JacksonMaxfield (approved):
This all looks good! I have a single question but otherwise ready for merge.

src/pages/EventPage/EventPage.tsx
tohuynh: >shares url?s=1 will that route to session 1 and default to 0 seconds?

Yes, but s=1 will route to session 2 (s starts at 0).

>shares the seconds url?t=100 will that default to session=0?

Yes

This only happens if the user alters the shareable URL though. The generated shareable URL will contain `s` and `t` (if start at checkbox is checked and t>0)

PR #157, opened 2021-12-31
feature/json-validation

### LINKED ISSUE #150: JSON Validation
### JSON Validation

The networking layer of CDP frontend currently does not validate the JSON it receives from the backend. This means that every property on every Model object is optional (denoted with a `?`).  

This feature is done when the incoming JSON is validated against the backend schema.  Therefore, every required property in the backend should be non-optional in the frontend. 

### Solution

Ideally we implement a function on the Model object (which every model extends), and then call that function prior to the model object's constructor. This may get a little bit sticky with TypeScript.

JSON validation will require a couple things:

- A function on the model.
- A list of required properties on every model.
- Calling the function to see if the JSON provided has the required properties (and throwing an error if the required properties are not present)

### Alternatives

If it were possible to have fail-able constructors in Javascript I think I might prefer that. However, JS constructors must return an object of some kind (an Error is possible!). If we had failable constructors I think it would be very annoying in the model layer.

BrianL3:
### Link to Relevant Issue

This pull request resolves #150 

### Description of Changes

As part of ModelService's sanity checks (did anything get returned from NetworkService?  was there an error? etc) before it creates a Model, it now checks to see if the required properties are on the json. This does not cascade down, but we have very flat database objects so it works out. 

This enables us to have required properties on our Model objects.

[Storybook (no changes to visual components)](https://{BrianL3}.github.io/cdp-frontend/)

### DIFF
--- a/src/models/Event.ts
+++ b/src/models/Event.ts
@@ -20,6 +20,9 @@ class Event implements Model {
 
   constructor(jsonData: ResponseData) {
     this.id = jsonData["id"];
+    console.log(`Attempting to create Event with data: ${JSON.stringify(jsonData, null, 2)}`);
+    this.event_datetime = firestoreTimestampToDate(jsonData["event_datetime"]);
+    this.body_ref = jsonData["body_ref"].id;
 
     if (jsonData["agenda_uri"]) {
       this.agenda_uri = jsonData["agenda_uri"];
@@ -29,13 +32,12 @@ class Event implements Model {
       this.minutes_uri = jsonData["minutes_uri"];
     }
 
-    this.body_ref = jsonData["body_ref"].id;
-
-    if (typeof jsonData["body_ref"] === "object") {
+    if (
+      typeof jsonData["body_ref"] === "object" &&
+      !(jsonData["body_ref"] instanceof DocumentReference)
+    ) {
       this.body = new Body(jsonData["body_ref"]);
     }
-
-    this.event_datetime = firestoreTimestampToDate(jsonData["event_datetime"]);
 
     if (jsonData["external_source_id"]) {
       this.external_source_id = jsonData["external_source_id"];
--- a/src/models/Event.ts
+++ b/src/models/Event.ts
@@ -6,11 +6,11 @@ import { Model } from "./Model";
 import { DocumentReference } from "@firebase/firestore";
 
 class Event implements Model {
-  id?: string;
+  id: string;
   agenda_uri?: string;
-  body_ref?: string;
+  body_ref: string;
   body?: Body;
-  event_datetime?: Date;
+  event_datetime: Date;
   external_source_id?: string;
   hover_thumbnail_ref?: string;
   hover_thumbnail?: File;
@@ -19,9 +19,7 @@ class Event implements Model {
   static_thumbnail?: File;
 
   constructor(jsonData: ResponseData) {
-    if (jsonData["id"]) {
-      this.id = jsonData["id"];
-    }
+    this.id = jsonData["id"];
 
     if (jsonData["agenda_uri"]) {
       this.agenda_uri = jsonData["agenda_uri"];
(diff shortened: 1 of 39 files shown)


src/models/TranscriptJson.ts
```diff
@@ -33,6 +34,8 @@ export default class TranscriptJson implements Model {
   annotations?: any;
```
tohuynh, line 34: TranscriptJson is not a database collection, it is just a JSON of the transcript data. https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/transcript_model.py

So, it would not have an `id`. `confidence`, `generator`, `created_datetime`, `sentences` are required fields.

src/models/Vote.ts
```diff
@@ -4,45 +4,37 @@ import Person from "./Person";
 import Event from "./Event";
 import Matter from "./Matter";
 import { Model } from "./Model";
-import { DocumentReference } from "@firebase/firestore";
-
 export default class Vote implements Model {
-  id?: string;
+  id: string;
   decision?: string;
```
tohuynh, line 9: `decision` is required.

src/models/util/validateResponseData.ts
```diff
@@ -0,0 +1,77 @@
+import { ResponseData } from "../../networking/NetworkResponse";
+import { COLLECTION_NAME } from "../../networking/PopulationOptions";
+
+function getRequiredProperties(forCollection: COLLECTION_NAME): string[] {
```
tohuynh, line 4: Missing `TranscriptJson ` case

tohuynh (commented):
Have some fixes for reqruied fields and `TransriptJson`.

And a general question regarding the creation of new Model out of a model_ref pattern:
```
if (typeof jsonData["body_ref"] === "object") {
      this.body = new Body(jsonData["body_ref"]);
```

To pass the two test cases, add the required fields for the example `Event` at 
https://github.com/BrianL3/cdp-frontend/blob/feature/json-validation/src/networking/test/ModelService.test.ts#L10
https://github.com/BrianL3/cdp-frontend/blob/feature/json-validation/src/networking/test/ModelService.test.ts#L18

`Promise.resolve(new NetworkResponse({ agenda_uri: "test", <add required fields> }))`

src/networking/TranscriptJsonService.ts
```diff
@@ -2,14 +2,19 @@ import { FirebaseConfig } from "../app/AppConfigContext";
 import TranscriptJson from "../models/TranscriptJson";
 
 import ModelService from "./ModelService";
+import { COLLECTION_NAME } from "./PopulationOptions";
 
 export default class TranscriptJsonService extends ModelService {
   constructor(firebaseConfig: FirebaseConfig) {
-    super("TranscriptJson", firebaseConfig);
+    super(COLLECTION_NAME.Transcript, firebaseConfig);
   }
```
tohuynh, line 10: Let's add TranscriptJson to `COLLECTION_NAME` and use that? Even though it's not a collection

tohuynh:
`getRequiredProperties` also need to be updated after changing the model files to have missing required fields as well.

src/models/TranscriptJson.ts
BrianL3: got it, I was wondering what they were doing in the models because I couldn't find it in the [database schema](https://councildataproject.org/cdp-backend/database_schema.html)

src/models/Event.ts
```diff
@@ -31,18 +29,14 @@ class Event implements Model {
...
-      } else if (typeof jsonData["body_ref"] === "object") {
-        this.body = new Body(jsonData["body_ref"]);
-      }
-    }
+    this.body_ref = jsonData["body_ref"].id;
 
-    if (jsonData["event_datetime"]) {
-      this.event_datetime = firestoreTimestampToDate(jsonData["event_datetime"]);
+    if (typeof jsonData["body_ref"] === "object") {
+      this.body = new Body(jsonData["body_ref"]);
```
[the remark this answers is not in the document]
BrianL3: Got it - the if-else that used to be in place is now replicated with a `if typeof === object && !(<the json thing> instanceof DocumentReference)`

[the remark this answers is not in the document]
tohuynh: That works

JacksonMaxfield (approved):
This all looks good to me. Once again, I cannot wait until we get a backend person / a dev ops person to write some fancy code gen for this type of stuff.

PR #160, opened 2022-01-03
fix/person-and-seat-picture-src

### LINKED ISSUE #153: Person Views' Pictures
### Describe the Bug

PersonPage is not receiving the correct data to render the person's picture.  It is not receiving the seat picture either.

### Reproduction

_Steps to reproduce the behavior:_

1. Go to PersonPage on a mobile- or tablet-sized screen.
2. You will never see a person's picture or seat picture. 

### Fix

The person's picture isn't showing up because person.picture.uri is not a valid value for img.src. The download URL is needed instead. See https://github.com/CouncilDataProject/cdp-frontend/blob/main/src/networking/EventSearchService.ts#L255 for an example. 

To get the seatPictureSrc to display, we need the download URL of the seat.image.uri, same solution as for the person picture.

tohuynh:
### Link to Relevant Issue

This pull request resolves #153 

### Description of Changes

Retrieve and display the person's picture src and seat's picture src.

Added a `getDownloadUrl` method to `NetworkService` to retrieve the URL.

----
I modified the data retrieved by the Person Page a bit to reduce the number of requests to firestore. The person page needs "Councilmember" roles with only `Seat` populated, the rest of the roles need just the `Body` populated. I separated out these two calls to avoid unnecessary population. I could add another `where` (where("title" != "Councilmember"))clause to `getPopulatedRolesByPersonId` so it only retrieve non-councilmember roles, but the seattle-staging DB was complaining about the composite index hasn't been created.

See [deployment](https://tohuynh.github.io/cdp-frontend/#/people/e54faa3434c9) for council member and seat picutres.

### DIFF
--- a/src/components/Cards/PersonCard/PersonCard.tsx
+++ b/src/components/Cards/PersonCard/PersonCard.tsx
@@ -31,7 +31,7 @@ export interface PersonCardProps {
   /** The person's name */
   personName: string;
   /** The person's picture */
-  personPictureSrc: string;
+  personPictureSrc?: string;
   /** Is the person a current councilmember? */
   personIsActive: boolean;
   /** The seat's name */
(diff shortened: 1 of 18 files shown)


JacksonMaxfield (approved):
> seattle-staging DB was complaining about the composite index hasn't been created

Would you mind making the index? https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L179

That collection has so many I am suprised we missed it.

Otherwise this looks good. Minor nitpick would be, if possible can you remove the border and background from the person image?

image

I think it looks a tiny bit better / my nitpicky-ness is fixed haha.

tohuynh:
> Otherwise this looks good. Minor nitpick would be, if possible can you remove the border and background from the person image?

I'm fine with that? What about you @BrianL3 ?

tohuynh:
> Would you mind making the index? https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L179

I tested the adddion of the second where clause, but firestore is forcing me to add `orderBy("title")` before `orderBy("start_datetime", ORDER_DIRECTION.desc)`, which I think we don't want. I'll just add the second where clause and leave out both orderBy clauses and sort the roles on the frontend (after the composite index has been created)? Seems alright?

PR #162, opened 2022-01-06
admin/remove-external-deps-from-bundle

tohuynh:
I'm not 100% sure about this. The cause of the error was because `moment.js` was included in our bundle (index.umd.js and index.es.js) which can include unresolvable imports like `./locale` (because it doesn't know it's supposed to find `locale` in `node_modules`.

I removed the external deps from our bundle by adding them to the `external` field in `vite.lib.config.js`. I don't know some of the names for these external modules (that should be added to `output.globals`). So I let vite guess their names based on the usage. It seems to work fine though.

I copied the content of `cdp-frontend/dist` into `king-county/web/node_modules/@councildataproject/cdp-frontend/dist` and then ran `start` and `build`. It seems to work fine.

### DIFF
--- a/vite.lib.config.js
+++ b/vite.lib.config.js
@@ -2,6 +2,8 @@ import path from "path";
 import { defineConfig } from "vite";
 import react from "@vitejs/plugin-react";
 
+import { dependencies } from "./package.json"
+
 export default defineConfig({
   plugins: [
     react({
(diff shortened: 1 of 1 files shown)


JacksonMaxfield (approved):
i always forget that you can just _import_ something from a json file in JS. neat!

PR #165, opened 2022-01-07
fix/cards-width

tohuynh:
### Link to Relevant Issue

This pull request resolves #

### Description of Changes

Change column's width to `1fr` so that it grows to fit the column width.

image


JacksonMaxfield (approved):
Nice!

PR #176, opened 2022-01-15
feature/better-bios

### LINKED ISSUE #151: Better Bios
### Biographies for People Could be Better

Biographies, as seen in the component <Biography> are implemented in an unfinished state. 

- They currently are not sensitive to the Person's current role.
- They are not localized.
- The sentence for number of terms served is not necessarily accurate.
- See line 40, filterRolesByTitle - We want to filter by ROLE_TITLE.COUNCILMEMBER here to get the number of roles for which the person is a council member in order to get how many terms the person has served. However, there aren't any roles in the DB with title == COUNCILMEMBER (this is probably a backend issue).
Or we can filter by role.body.name == "City Council" to get the correct number of roles to determine the number of terms.
But both methods of determining how many terms the person has served isn't accurate if the backend doesn't backfill data. For example, if the backend starts collecting data only recently the number of terms could be inaccurate.

I think what we want to do is break down all possible Roles and have custom helper sentence constructors for each.  The writeBiography function can then distinguish between roles that the Person has had in the past, is currently serving, etc, and be much more robust.

Perhaps the most important part of this feature is testing.

### Use Case

Once we have a more robust writeBiography in place, every Person no matter their current or past role will have a biography generated that makes sense.

### Solution

Done When:

- Fix the things listed under the top section.
- Storybook has bios for every role/person type including edge cases.

### LINKED ISSUE #158: Biography Fails when Deals with a Person without a Current Role
### Describe the Bug

When a Person does not have a current role, the PersonPage either crashes or has other problems.  I have an incoming PR that will fix the crashes, but the problems with the biography having a bunch of undefined stuff remains.

### Reproduction

_Steps to reproduce the behavior:_

1. Go to the People Page
2. Click on Daniel Strauss
3. See bad biography (due to not holding any current role I think?)

Instead of 

### Screenshots

Screen Shot 2021-12-31 at 5 25 10 PM

BrianL3:
### Link to Relevant Issue

This pull request resolves #158  (this bug is no longer applicable, as People must have at least Councilmember roles according to @JacksonMaxfield )
This pull request also resolves #151 request for better biography information on the PersonPage, and better mocks.

### Description of Changes

This is a simple change to how Bios look.  It provides somewhat more detailed information about a Councilperson.

### Link to Forked Example App
Changes are visible on individual's PersonPage(s)
https://BrianL3.github.io/cdp-frontend/

### DIFF
--- a/src/models/util/RoleUtilities.ts
+++ b/src/models/util/RoleUtilities.ts
@@ -1,6 +1,7 @@
 import Role from "../Role";
 
 enum ROLE_TITLE {
+  ALTERNATE = "Alternate",
   CHAIR = "Chair",
   VICE_CHAIR = "Vice Chair",
   COUNCILMEMBER = "Councilmember",
@@ -12,6 +13,51 @@ function filterRolesByTitle(roles: Role[], title: ROLE_TITLE): Role[] {
   return roles.filter((role) => {
     return role.title === title;
   });
+}
+
+function getUniqueBodyRoles(roles: Role[]): Role[] {
+  const setOfBodies: any = {};
+  return roles.filter((role) => {
+    if (role.body?.id) {
+      if (setOfBodies[role.body.id] === true) {
+        return false;
+      } else {
+        setOfBodies[role.body.id] = true;
+        return true;
+      }
+    }
+  });
+}
+
+function getCurrentUniqueBodyRoles(roles: Role[]): Role[] {
+  return getUniqueBodyRoles(roles)
+    .filter((role) => {
+      if (role.title === ROLE_TITLE.COUNCILMEMBER) {
+        return false;
+      }
+      if (!role.end_datetime) {
+        return true;
+      } else {
+        if (role.end_datetime > new Date()) {
+          return true;
+        } else {
+          return false;
+        }
+      }
+    })
+    .sort((roleA: Role, roleB: Role) => {
+      if (roleA.title === ROLE_TITLE.CHAIR) {
+        return -1;
+      }
+      if (roleA.title === ROLE_TITLE.VICE_CHAIR && roleB.title !== ROLE_TITLE.CHAIR) {
+        return -1;
+      }
+      if (roleA.title === ROLE_TITLE.ALTERNATE && roleB.title !== ROLE_TITLE.MEMBER) {
+        return -1;
+      }
+
+      return 1;
+    });
 }
 
 function getMostRecentRole(roles: Role[]): Role {
@@ -38,4 +84,4 @@ function getMostRecentRole(roles: Role[]): Role {
   return roles[0];
 }
 
-export { ROLE_TITLE, filterRolesByTitle, getMostRecentRole };
+export { ROLE_TITLE, filterRolesByTitle, getCurrentUniqueBodyRoles };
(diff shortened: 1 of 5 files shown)


tohuynh:
Some thoughts from looking at the example app:
- The sentence like `Councilmember González represents Citywide as Councilmember of Position 9` needs to reflect whether the person is active or not. (Also add `active` to the heading element, next to the councilmember name, like https://councildataproject.org/cdp-frontend/?path=/story/library-cards-person--with-seat-picture ?) Maybe also display the start_datetime and end_datetime, since we have them.
- The chair sentence like `They currently serve as Chair of the following bodies: Economic Development, Technology, and City Light Committee` is not needed since we list the chairs in the `<ul>`.
- I think we want to display all roles, not just active ones. Maybe display something along the line of 
```
bio using the most recent councilmember role 
list previous council member roles (with seat, start_datetime, end_datetime info)
list all non-councilmember roles (chair, vice-chair, etc) with active ones first ( specify that active ones with `active` or add start_datetime and end_datetime next to the body names)
```
?

---
Thoughts beyond the bio:
- Since we got matters sponsored and will later have the matter page, let's display a section of links for each matter sponsored?

JacksonMaxfield:
Hmmmmm a part of me is realizing this bio thing is a bigger problem as you raise To. Like the role history should maybe be a table, same with the "sponsored legislation"? Our designer dropped off iirc. I wonder if Neha would be interested in working on this page...

Strictly for the bio. I agree with To, the active badge should be used:
"Councilmember Gonzalez represented..." for inactive and "represents" for active.

The sentence flow is a bit odd with _just_ the region (Citywide) we could make it a bit better with: "Councilmember Gonzalez represented Citywide residents as the councilmember for Position 9." Or maybe, "Councilmember Gonzalez served as the the councilmember for Position 9 representing Citywide residents."

Similarly the: "They currently hold no Chairs" could maybe be changed to: "They currently do not chair any committees."

But I am starting to think maybe this bio thing is just crazy. I feel like this entire thing could be turned into a quick like bullet list near the contact info or something. Really want to ask Neha or other designers for input.

tohuynh:
The more I look at this, I think we should move away from writing sentences because it's hard to get right and localize.

I think something like this could work for CDP
image

At the top of the page, we have
```
Person name (inactive/active badge) (or add the start_datetime and/or end_datetime somewhere)
Seat name -- seat electoral area
seat picture and person picture
```

```
Previous term:
  <list of previous terms>
```

```
Committees (non-councilmember roles):
  <list of active roles>
  <list of inactive roles, initially hidden like accordion maybe?>
```

```
Matter Sponsored:
  <list of matters sponsored>
```

tohuynh:
> Really want to ask Neha or other designers for input.

Agree

src/models/util/RoleUtilities.ts
```diff
@@ -14,6 +15,51 @@ function filterRolesByTitle(roles: Role[], title: ROLE_TITLE): Role[] {
...
+      if (roleA.title === ROLE_TITLE.ALTERNATE && roleB.title !== ROLE_TITLE.MEMBER) {
+        return -1;
+      }
+
+      return 1;
+    });
+}
+
 function getMostRecentRole(roles: Role[]): Role {
   roles.sort((a, b) => {
```
tohuynh, line 64: `getMostRecentRole` isn't used anymore right? If yes, we can remove it.

tohuynh (approved):
Overall, the PR added better mocks and is an improvement from the previous person page.

I'm fine with merging even if we decide to change the person page later on.

JacksonMaxfield (approved):
Agree. I think we simply need a design pass over all of this. Merge whenever.

PR #182, opened 2022-01-20
admin/update-db-model-constants

tohuynh:
### Link to Relevant Issue

This pull request resolves #

### Description of Changes

cdp-frontend `VOTE_DECISION` is out of date with cdp-backend `VoteDecision`

- Update `VOTE_DECISION` with backend's [vote decision](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/constants.py#L15)
- Move the db constants into models/constants.ts. I think it makes more sense to place them there
- Localise `VOTE_DECISION` (they still need to be translated to de and es, will make a general issue for that)
- Change `<number> Abstained` to `<number> Non-Voting` in the MeetingVotesTable ("Non-Voting" also need to be translated)
image
- Change the field type in the model files to their appropriate types. e.g. eventMinutesItem.decision: EVENT_MINUTES_ITEM_DECISION. I didn't write any validation for this conversion because the backend already does [that](https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L559).
- Clean up a few table stories so that the example data are auto-generated or have the correct types

### DIFF
--- a/src/assets/LocalizedStrings.tsx
+++ b/src/assets/LocalizedStrings.tsx
@@ -39,10 +39,17 @@ export interface MasterStringsList extends LocalizedStringsMethods {
   tools_license_prefix: string;
   styled_using_prefix: string;
   artwork_provided_by_prefix: string;
+  passed: string;
+  failed: string;
   approve: string;
   reject: string;
+  abstain_approve: string;
+  abstain_non_voting: string;
+  abstain_reject: string;
+  absent_approve: string;
+  absent_non_voting: string;
+  absent_reject: string;
   rejected: string;
-  abstain: string;
   adopted: string;
   in_progress: string;
   example_prefix: string;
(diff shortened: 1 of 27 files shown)


JacksonMaxfield (approved):
Looks good to me. I thought we were planning on removing german translations entirely because no one on the team speaks german :joy: 

Also, we should add the localization widget to the actual app somewhere so these can actually be used.

tohuynh:
> Also, we should add the localization widget to the actual app somewhere so these can actually be used.

Yes, for sure, after all texts are properly translated and added to our localization feature

tohuynh:
> I thought we were planning on removing german translations entirely because no one on the team speaks german 😂

Don't we have a german person now (Pia)?

JacksonMaxfield:
> > I thought we were planning on removing german translations entirely because no one on the team speaks german joy
> 
> Don't we have a german person now (Pia)?

We do indeed.

PR #185, opened 2022-01-21
fix/temp-remove-matter-links

tohuynh:
Temporarily replaces the matter links(that leads to the matter page that doesn't exist atm) with `p` that is bolded.

image


JacksonMaxfield (approved):
Thank you!

PR #192, opened 2022-02-08
component/legislation-introduction

### LINKED ISSUE #172: [f_legislationtracking] Legislation introduction component
Implement the Legislation introduction component of the legislation details page (`/matters/:id`).

See the figma lofi mocks ups https://www.figma.com/file/VPeHF2SVDn6FneLLl9bihB/Legislation-Tracking-(CDP)?node-id=137%3A350
image

- The legislation name (e.g. CB 120237) should use an `h1` with `className=mzp-u-title-sm`
- The bill status(In committee) should use `MATTER_STATUS_DECISION` constant and the corresponding icons. The three icons are Adopted, Rejected, In Progress.
- @JacksonMaxfield After the bill status, are we going to display a list of topics(we don't have this data?) or keywords(we will have this data after the matter index pipeline) ?

BrianL3:
### Link to Relevant Issue

This pull request resolves #172 

### Storybook

See the component [here](https://brianl3.github.io/cdp-frontend/?path=/story/library-details-legislation-introduction--approval-story

### DIFF
--- a/src/components/Details/Legislation/LegislationIntroduction.tsx
+++ b/src/components/Details/Legislation/LegislationIntroduction.tsx
@@ -1,0 +1,53 @@
+import React, { FC } from "react";
+import MatterStatus from "../../../models/MatterStatus";
+import IndexedMatterGram from "../../../models/IndexedMatterGram";
+import { DecisionResult } from "../../Shared";
+
+export interface LegislationIntroductionProps {
+  matterStatus: MatterStatus;
+  indexedMatterGrams: IndexedMatterGram[];
+}
+
+const LegislationIntroduction: FC<LegislationIntroductionProps> = ({
+  matterStatus,
+  indexedMatterGrams,
+}: LegislationIntroductionProps) => {
+  return (
+    <div>
+      <h1 className="mzp-u-title-sm">{matterStatus.matter?.name}</h1>
+      <b className="mzp-u-body-sm">{matterStatus.matter?.title}</b>
+      <div
+        style={{
+          display: "flex",
+          flexDirection: "row",
+          justifyContent: "start",
+          marginTop: 12,
+          marginBottom: 12,
+        }}
+      >
+        <DecisionResult
+          result={matterStatus.status}
+          style={{ display: "flex", alignItems: "center", flex: 0, whiteSpace: "nowrap" }}
+        />
+        <div
+          style={{
+            display: "flex",
+            flexDirection: "row",
+            justifyContent: "left",
+            alignItems: "center",
+            flex: 1,
+          }}
+        >
+          {indexedMatterGrams.map((indexedMatterGram) => (
+            <div key={indexedMatterGram.id}>
+              <p style={{ display: "inline", paddingLeft: 11, paddingRight: 11 }}>•</p>
+              <p style={{ display: "inline" }}>{indexedMatterGram.unstemmed_gram}</p>
+            </div>
+          ))}
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default LegislationIntroduction;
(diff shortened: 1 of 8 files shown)


src/components/Details/Legislation/LegislationIntroduction.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const LegislationIntroduction: FC<LegislationIntroductionProps> = ({
+  matterStatus,
+  indexedMatterGrams,
+}: LegislationIntroductionProps) => {
+  return (
+    <div>
+      <h1 className="mzp-u-title-sm">{matterStatus.matter?.name}</h1>
+      <b className="mzp-u-body-sm">{matterStatus.matter?.title}</b>
+      <div
```
tohuynh, line 19: I don't think `mzp-u-body-sm` is valid Mozilla protocol CSS class

src/components/Shared/DecisionResult.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import React, { CSSProperties, FC, StyleHTMLAttributes } from "react";
```
tohuynh, line 1: `StyleHTMLAttributes` can be removed

src/models/IndexedMatterGram.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (
+      typeof jsonData["matter_ref"] === "object" &&
+      !(jsonData["matter_ref"] instanceof DocumentReference)
+    ) {
+      this.matter = new Matter(jsonData["event_ref"]);
+    }
+
+    if (jsonData["keywords"]) {
+      this.keywords = jsonData["keywords"];
+    }
```
tohuynh, line 35: `IndexedMatterGram` are the keywords, at least the top 5, of a `Matter`. I don't think `.keywords` would be a field on `IndexedMatterGram`.

cc @JacksonMaxfield

tohuynh (approved):
Looks good. Have a few minor asks.

The addition to package-lock.json seems innocuous, but why was it added? I don't see any changes to package.json

src/models/IndexedMatterGram.ts
JacksonMaxfield: Yes the "keywords" prop should be removed but to be fair, looks like the mocked models dont implement the keywords prop :shrug: 

My opinion is its okay to leave as is because after I finish the matter index pipeline I can come update this class too.

JacksonMaxfield (approved):
Makes sense to me! Sorry that the matter index pipeline isn't done yet.

src/components/Details/Legislation/LegislationIntroduction.tsx
BrianL3: where can I find a list of valid classes?  I use their website but in a case like this where I want something that's just a little bit different from the listed items, it becomes very difficult to figure out what is valid and what is not.

src/models/IndexedMatterGram.ts
BrianL3: Right - after talking with Jackson yesterday, it seems the "keyword" is the unstemmed gram property. The "keywords" list then, is simply a list of the unstemmed grams from an array of IndexedMatterGrams.

Removed this wrong property.

src/components/Details/Legislation/LegislationIntroduction.tsx
tohuynh: Not sure if there's a page that lists them. I didn't find the class inside `node_modules/@mozilla-protocol/core/css`

tohuynh: https://protocol.mozilla.org/docs/naming.html

Here's how they name their class though

PR #197, opened 2022-02-10
bugfix/localization-updates

### LINKED ISSUE #183: Localization effort
This an issue to keep track of which texts need to be provide translations and added to our localization feature

### texts that need translations
- text in LegislationOverview
- text in LegislationLatestVote

### texts that need to be localized (add to LocalizedStrings.tsx and the supported lang files)

@JacksonMaxfield @BrianL3 @tohuynh 
This is an on-going story, just change the text of this box as needed.

BrianL3:
### Link to Relevant Issue

This pull request resolves #183 

### Changes

Adding many new strings to the localized files. We do not yet have a firm idea on placement of the localization widget. However, getting strings localized as we go will be of great help when we finally do add the widget.

### DIFF
--- a/src/assets/LocalizedStrings.tsx
+++ b/src/assets/LocalizedStrings.tsx
@@ -1,6 +1,5 @@
 import * as Languages from "./strings";
 import LocalizedStrings, { LocalizedStringsMethods } from "react-localization";
-import { StringNullableChain } from "lodash";
 /**
  * When adding strings:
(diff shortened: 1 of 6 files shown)


tohuynh:
Looks good, but let's keep the gh-issue open and update it if there's any text that still needs translating and added to the localization feature?

tohuynh (approved):

PR #199, opened 2022-02-12
component/legislation-latest-vote

### LINKED ISSUE #174: [f_legislationtracking] Legislation latest vote component
Implement the Legislation latest vote component of the legislation details page (`/matters/:id`).

See the figma lofi mocks ups https://www.figma.com/file/VPeHF2SVDn6FneLLl9bihB/Legislation-Tracking-(CDP)?node-id=137%3A350

image

- The vote (Yay, Nay, etc) should use our `VOTE_DECISION` constants
- The council member should link to the `/people/:id` page

BrianL3:
### Link to Relevant Issue

This pull request resolves #174 

### Description of Changes

A draft of the Latest Vote component for the Legislation details page.

[Storybook](https://brianl3.github.io/cdp-frontend/?path=/story/library-details-legislation-latestvote--adopted)


src/components/Details/Legislation/VoteBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  flex: 1,
+  justifyContent: "space-between",
+  alignItems: "center",
+  zIndex: props.zIndex,
+}));
+
+const VOTE_DECISION_COLOR = {
+  [VOTE_DECISION.APPROVE]: "cdp-bg-acceptance-green",
+  [VOTE_DECISION.ABSTAIN_APPROVE]: "cdp-bg-acceptance-green",
+  [VOTE_DECISION.ABSENT_APPROVE]: "cdp-bg-acceptance-green",
```
JacksonMaxfield, line 31: these triple bars with the same colors may get a bit confusing but I get it for now.

src/components/Details/Legislation/LegislationLatestVote.tsx
```diff
@@ -0,0 +1,75 @@
...
+      key={"againstVoteBar"}
+      status={VOTE_DECISION.REJECT}
+      votes={against}
+      percentage={(against.length / votes.length) * 100}
+      height={VOTE_BAR_HEIGHT}
+      label={votesAgainstLabel}
+    />,
+    <VoteBar
+      key={"abstainedVoteBar"}
+      status={VOTE_DECISION.ABSTAIN_NON_VOTING}
```
JacksonMaxfield, line 51: Remind me, each muni can have different abstaining rules correct? Like some munis have abstains count as approve, some count abstain as reject, and then there are the "true abstain" like "i was in surgery and literally couldn't vote".

We are defaulting to using the third option here correct? In the future should this be customizable by muni?

JacksonMaxfield (commented):
Just assigned reviewers and assigned you as owner. But I think our discussion in Slack left us with some minor changes requested which were:

1. bar labels (Approve, Reject, Abstain, etc) should always be on the left, the actual number count should be on the right side of the bar
2. investigate a plotting library over using divs? (this is less important) -- especially because of the hover interactions that come with existing plotting tools?

src/components/Details/Legislation/LegislationLatestVote.tsx
BrianL3: The status prop is really just used for color.  There will always be _only_ 3 bars: green/red/purple. Those three bars will merge all the relevant votes into themselves. E.g. There are 3 votes: VOTE_DECISION.ABSENT_APPROVE, VOTE_DECISION.ABSTAIN_APPROVE, VOTE_DECISION.APPROVE: they are ALL merged into the green bar. 

I prefer this over have a bunch of different bars which really mean the same thing (3 different green bars in the above example). If the user feels the need to see the _exact_ vote decision, we could have it in the expanded view).

src/components/Details/Legislation/VoteBar.tsx
BrianL3: There will never be 3 green bars, only 1. What this does is merge the 3 "approve" type votes into 1 green bar.

src/components/Details/Legislation/LegislationLatestVote.tsx
JacksonMaxfield: I agree that this is preferred than having 9 different bars. But this is a _perfect_ use case for the horizontal stacked bar. Being able to differentiate between different vote types _that added to bill approval_. I.e. one part of the bar is "true approve" and another part of the bar is "abstain (approve)" for example but they get slotted into the same bar.

src/components/Details/Legislation/VoteBar.tsx
BrianL3: I clarified the code to ensure no confusion. A simple color is passed in.

src/components/Details/Legislation/LegislationLatestVote.tsx
BrianL3: I added the exact vote decision as part of the expanded view so there can be no confusion.

src/components/Details/Legislation/VoteBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+  height: props.height,
+  width: props.width,
+  borderTopLeftRadius: "50em",
+  borderBottomLeftRadius: "50em",
+  borderTopRightRadius: "50em",
+  borderBottomRightRadius: "50em",
+  display: "flex",
+  flex: 1,
+  justifyContent: "space-between",
+  alignItems: "center",
```
tohuynh, line 23: display, flex, justifyContent, alignItems can be removed because ColorBar doesn't have any children.

tohuynh (commented):
Regarding the bar graph and the placement of the list of council members, I think we should match Neha's design as much as possible since it went through some user testing, though there were only 4 users.

I think the bar graph should stay unchanging so users can view the bar graph as a whole while viewing who approved/rejected/abstained. (Clicking on the individual bar causes the bar graph to look different).

src/components/Details/Legislation/VoteBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+interface VoteBarProps {
+  statusColor: string;
+  votes: Vote[];
+  percentage: number;
+  height: number;
+  label: string;
+}
+
+const ProgressBar: FC<VoteBarProps> = ({
+  statusColor,
```
tohuynh, line 35: ProgressBar -> VoteBar

src/components/Details/Legislation/VoteBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <ColorBar className="cdp-bg-neutral-grey" width={`100%`} height={height} zIndex={1} />
+        </div>
+        <b style={{ marginLeft: 16, width: textSize }}>{label}</b>
+      </div>
+      {expanded && votesLinks}
+    </div>
+  );
+};
+
+export default ProgressBar;
```
tohuynh, line 87: ProgressBar -> VoteBar

BrianL3:
Let's iterate from here. The component as currently written works for mobile, long localized strings, and other weirdness. Improvements and redesigns can be thrown on from here.

Issue #166, opened 2022-01-10
Allow the `/people/:id` page to render before person's votes and/or biography data is fetched.


tohuynh:
### Feature Description

_A clear and concise description of the feature you're requesting._

Similar to #101.
Allow the `/people/:id`page  to render before a person's votes and/or biography data is fetched. Depending on how fast the request for biography data is fulfilled, we might need to lazy load the biography data too.

### Use Case

_Please provide a use case to help us understand your request in context._

Reduce the page's "time to interactivity".  Users don't need to see a person's votes right away.

### Solution

_Please describe your ideal solution._

Only load things as they come in. I.e. Maybe the voting table sections show a "Fetching votes..." line until they have returned?

See /src/pages/EventPage for starting point.

JacksonMaxfield:
Neha (UX designer working on legislation page) had a nice idea of putting stuff below the fold and only rendering / fetching when clicked.

![image](https://user-images.githubusercontent.com/17132317/148819757-40fc5526-3670-4a9d-8247-da6e3b61c8f6.png)

The "History" portion of the page is collapsed on first load and background loaded. Data is revealed when accordion is opened. We may be able to do the same thing with the votes.

tohuynh:
Seems like a good idea. Would be good to have consistent UI across pages.

PR #203, opened 2022-02-22
feature/optimize-person-page

### LINKED ISSUE #194: Allow the person page to load before the list of matter sponsored is fetched
### Feature Description

_A clear and concise description of the feature you're requesting._

Similar to #101 and #166 .
Allow the `/people/:id`page  to render before a person's matter sponsored data is fetched.

### Use Case

_Please provide a use case to help us understand your request in context._

Reduce the page's "time to interactivity".  Users don't need to see a person's matter sponsored right away.

### Solution

Only load things as they come in. I.e. Maybe the section show a "Fetching legislation sponsored by "person name"..." line until they have returned?

See /src/pages/EventPage for starting point.

tohuynh:
### Link to Relevant Issue

This pull request resolves #194 #166 

### Description of Changes

Reduce the "time to interactivity" of the person page by allowing it to render before
- the person and seat pictures are loaded (the real culprit of our current and slow person page)
- the matter sponsored by a person are fetched
- the votes of the person are fetched

The approach is the same as before for the event page. Separate each call to the DB into it's own `useFetchData` call, then decide which data is needed for the page to render. For the person page, it's the `person` and their `roles`.

I added a little component for displaying the "loading" skeleton (sort of a placeholder UI) for pictures so that the layout of the page doesn't change when the pictures are loaded.

Created `LazyFetchDataContainer`, a non-blocking version of `FetchDataContainer` to handle the rendering logic -- is it fetching data, is there an error, and to display the data when it is fetched. Although these two containers look similar, code-wise, and could be combined into one, I decided to keep them separated because the rendering logic for pages and smaller components could diverge in the future.

https://tohuynh.github.io/cdp-frontend/#/people/6410ce724625

### DIFF
--- a/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
+++ b/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
@@ -8,6 +8,8 @@ import { initialFetchDataState } from "../../../containers/FetchDataContainer/us
 
 import TranscriptSearch, { TranscriptSearchProps } from "./TranscriptSearch";
 
+import { mockSentences } from "../../../stories/model-mocks/sentence";
+
 export default {
   component: TranscriptSearch,
   title: "Library/Details/Transcript Search",
@@ -27,27 +29,7 @@ Default.args = {
   searchQuery: "sentence",
   sentences: {
     ...initialFetchDataState,
-    data: [
-      {
-        session_index: 0,
-        index: 0,
-        start_time: 1.5,
-        text: "This is a sentence.",
-        speaker_index: 0,
-        speaker_name: "Lisa Herbold",
-        speaker_id: "lisa-herbold",
-        speaker_pictureSrc:
-          "https://www.seattle.gov/images/Council/Members/Herbold/Herbold_225x225.jpg",
-      },
-      {
-        session_index: 0,
-        index: 1,
-        start_time: 2,
-        text: "This is another sentence.",
-        speaker_name: "Unknown",
-        speaker_index: 1,
-      },
-    ],
+    data: mockSentences(1, 2),
   },
   jumpToVideoClip: (sessionIndex, startTime) => action("jump to video clip"),
   jumpToTranscript: (sentenceIndex) => action("jump to transcript"),
@@ -58,17 +40,7 @@ manySentences.args = {
   searchQuery: "sentence",
   sentences: {
     ...initialFetchDataState,
-    data: Array.from({ length: 200 }).map((_, i) => ({
-      session_index: 0,
-      index: i,
-      start_time: i,
-      text: `This is a sentence ${i}.`,
-      speaker_name: "Lisa Herbold",
-      speaker_id: "lisa-herbold",
-      speaker_index: 0,
-      speaker_pictureSrc:
-        "https://www.seattle.gov/images/Council/Members/Herbold/Herbold_225x225.jpg",
-    })),
+    data: mockSentences(2, 100),
   },
   jumpToVideoClip: (sessionIndex, startTime) => action("jump to video clip"),
   jumpToTranscript: (sentenceIndex) => action("jump to transcript"),
--- a/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
+++ b/src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
@@ -4,6 +4,8 @@ import React from "react";
 import { action } from "@storybook/addon-actions";
 import { Story, Meta } from "@storybook/react";
 
+import { initialFetchDataState } from "../../../containers/FetchDataContainer/useFetchData";
+
 import TranscriptSearch, { TranscriptSearchProps } from "./TranscriptSearch";
 
 export default {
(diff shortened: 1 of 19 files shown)


src/containers/PersonContainer/CoverImage.tsx
```diff
@@ -61,25 +68,34 @@ const CoverImage: FC<CoverImageProps> = ({
...
-      />
-      {personPictureSrc && (
+      <PlaceholderWrapper contentIsLoading={seatPictureSrcIsLoading || personPictureSrcIsLoading}>
+        <Seat
+          className="mzp-c-card-image"
+          src={seatPictureSrc.data === null ? EXAMPLE_COVER_VIEW : seatPictureSrc.data}
+          alt={`${electoralArea || ""}`}
+          onLoad={() => setSeatPictureSrcIsLoading(false)}
+        />
+      </PlaceholderWrapper>
```
JacksonMaxfield, line 91: I see this placeholder but I also see you are using a fetch data container instead of the `LazyFetchDataContainer` for these values. Is the placeholder worth it then?

JacksonMaxfield (approved):
This all looks good to me. There is a lot so hopefully I didn't miss anything but the forked website looks like everything is running properly so :shrug: :shrug: 

I have one minor comment in the review but I have one more general comment which is: if the pictures are the items that are taking the longest to load, is it because of their URI fetching or is it because of their content fetching? i.e. should we downsample the images during storage?

src/containers/PersonContainer/CoverImage.tsx
tohuynh: `FetchDataContainer` is for containers that require lots of data from different DB models like EventContainer, PersonContainer, while `LazyFetchDataContainer` is for containers that require data from one or two DB models, like the components for votes of a person, or legislation sponsored by a person.

The `Placeholder` and the `LazyFetchDataContainer` are on the same level -- for use with smaller components. They are both needed. The placeholder is used to render a rectangular skeleton (in place of the actual pictures) so that the person name, contact, terms don't change positions when the pictures are finally loaded. While `LazyFetchDataContainer` is meant components that are initially hidden (no need to show rectangular skeleton).

tohuynh:
> I have one minor comment in the review but I have one more general comment which is: if the pictures are the items that are taking the longest to load, is it because of their URI fetching or is it because of their content fetching? i.e. should we downsample the images during storage?

They both contribute to the longer load time. I notice that the seat picture takes longer than the person's picture for the browser to load. And if the `file` could have a displayable URI (instead of gs://...), that would improve load time too.

tohuynh:
> There is a lot so hopefully I didn't miss anything but the forked website looks like everything is running properly so 🤷 🤷

There are 5-6 files where major changes happen. Most of it was just fixing storybook stories or incorporating `LazyFetchDataContainer` into the event page.

src/containers/PersonContainer/CoverImage.tsx
JacksonMaxfield: Makes sense to me! Thanks!

JacksonMaxfield:
> They both contribute to the longer load time. I notice that the seat picture takes longer than the person's picture for the browser to load. And if the `file` could have a displayable URI (instead of gs://...), that would improve load time too.

I think adding / storing an https URI to the `File` collection would be fine. Mind making an issue on `cdp-backend` for that. Granted, from my understanding, that extra query isn't the main thing slowing it down, the main thing is just the size of the image which to me says downsampling the images may help.

tohuynh:
I think the extra `getDownloadUrl` itself isn't so bad, but the load time (from old person page) adds up in fetch sequences -- e.g. get person, get file, get download URL, browser fetch content from URL.

Agree on the downsampling of the images would reduce the time it takes for the browser to retrieve the content.

src/containers/EventContainer/EventContainer.tsx
```diff
@@ -103,8 +103,8 @@ const EventContainer: FC<EventContainerProps> = ({
 
   //Create the TranscriptItemsRefs to create jumpToTranscript callback
   const transcriptItemsRefs = useMemo(() => {
```
BrianL3, line 105: Is the use of `useMemo` here so that if a component goes off screen and then returns, the cached version is used?

src/pages/PersonPage/PersonPage.tsx
```diff
@@ -1,86 +1,169 @@
...
-      personPictureSrc,
-      seatPictureSrc,
-      councilMemberRoles,
-    };
+  const fetchMatterSponsored = useCallback(async () => {
+    const matterSponsorService = new MatterSponsorService(firebaseConfig);
+    const mattersSponsored = await matterSponsorService.getMattersSponsoredByPersonId(id);
+    return Promise.resolve(mattersSponsored);
+  }, [id, firebaseConfig]);
+  const { state: mattersSponsoredDataState } = useFetchData<MatterSponsor[]>(
```
BrianL3, line 111: I don't understand JS syntax that well - what exactly is going on in this line? You're setting _something_ on the component's state, but I am not clear what or how.

BrianL3 (approved):
I like the addition of LazyLoadingContainers. I am not super sure I understand why they are being integrated directly into components, I was thinking they would be like a higher order component that is either inherited from, or injected into, the components that need it. I guess it makes sense since these are specifically-sized items that have to place-hold as their "real" components come in? 
I'm looking forward to learning more about this pattern.

src/components/Details/TranscriptSearch/TranscriptSearch.stories.tsx
```diff
@@ -23,45 +25,51 @@ const Template: Story<TranscriptSearchProps> = (args) => <TranscriptSearch {...a
...
-      index: 1,
-      start_time: 2,
-      text: "This is another sentence.",
-      speaker_name: "Unknown",
-      speaker_index: 1,
-    },
-  ],
+  sentences: {
+    ...initialFetchDataState,
+    data: [
```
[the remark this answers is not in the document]
tohuynh: Mock the `data` part? sounds good to me.

src/components/Shared/PlaceHolder.tsx
```diff
@@ -0,0 +1,67 @@
...
+  100% {
+    transform: translateX(100%);
+  }
+`;
+
+const Placeholder = styled.div<{ isLoading: boolean }>((props) => ({
+  position: "relative",
+  width: "100%",
+  height: "100%",
+  "& > div, & > div:nth-of-type(2):after": {
```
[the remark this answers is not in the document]
tohuynh: `&` I think is a feature of css-in-js library like emotion/styled. Everything else is just CSS. The `&` just refer to itself. `& > div` selects the div child of this element.

https://emotion.sh/docs/styled#nesting-components

src/pages/PersonPage/PersonPage.tsx
tohuynh: Just renaming `state` to `mattersSponsoredDataState `

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
go to "Assigning to new variable names"

tohuynh: `useFetchData` returns an object that has a field `state`. I'm just renaming the variable.

src/containers/EventContainer/EventContainer.tsx
tohuynh: If the value in the dependency array doesn't change, then it uses the cached version. This avoids re-calculation of the refs if it isn't needed (because the dependency array didn't change).

If the component gets destroyed then the value will have to be re-caclculated.

src/components/Shared/PlaceHolder.tsx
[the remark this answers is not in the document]
tohuynh: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

for `:after` `:nth-of-type` etc

tohuynh:
> I like the addition of LazyLoadingContainers. I am not super sure I understand why they are being integrated directly into components, I was thinking they would be like a higher order component that is either inherited from, or injected into, the components that need it. I guess it makes sense since these are specifically-sized items that have to place-hold as their "real" components come in?
I'm looking forward to learning more about this pattern.

It's sorta like a HOC, in that sense that it wraps around the component and add this extra rendering logic layer on top of the component. If the data is loading, display a message. If there is an error, display the error message. If the data is fetched, display the component that is meant to be displayed with data.

PR #205, opened 2022-03-08
feature/app-localization

tohuynh:
### Link to Relevant Issue

This pull request resolves #

### Description of Changes

More work toward the app localization feature.

I lifted the selected `language` state to the app level from the localization widget so that it be can use throughout the app. Gains include localized video player controls and date strings. The `language` and a dispatch to change the selected language are stored in a React context, like the app config context.

https://tohuynh.github.io/cdp-frontend/#/
(This is for demonstration only, the placement of localization widget in the footer is not included in the PR)


JacksonMaxfield (approved):
Wow, cool to actually see it work!

Only nitpick is the location of the language selector:
image

Can we move it to inline with the text / padded.

JacksonMaxfield (commented):
We may also want to add a little disclaimer that says like "Search, documents, and transcript will all remain in English"?

We don't want to say like this will update all elements include the transcript and search and stuff to work in other languages?

tohuynh:
> Only nitpick is the location of the language selector:

It's not included in the PR

JacksonMaxfield:
Woops. I should've read the PR closer :joy:

tohuynh:
> Woops. I should've read the PR closer 😂

No worries. This PR just says our localization feature will work when we add the widget.

Also fixes video.js lang support error.

src/containers/EventContainer/EventInfoTabs.tsx
```diff
@@ -96,6 +99,7 @@ const EventInfoTabs: FC<EventInfoTabsProps> = ({
   }, [eventMinutesItems, votesByEventMinutesItem]);
 
   const infoTabPanes = useMemo(() => {
+    console.log(`calculating tabs for language ${language}`);
```
BrianL3, line 102: Might want to remove the console logs if you're done with them.

BrianL3 (approved):
This is wonderful, To. I really like the architecture you've employed here. It allows for the usage of the same context across all the areas its needed: video requests, strings, localization of dates, etc.

src/containers/EventContainer/EventInfoTabs.tsx
tohuynh: Yeah, I wanted to, but it would cause a linting warning. "Remove `language` from the dependency array if it's not being used." But at the same time, I need `language` as a dependency because I wanted the tabs to be re-calculated whenever `language` changes.

PR #206, opened 2022-03-09
feature/paginate-votes

### LINKED ISSUE #156: Paginate the person's voting records
### Feature Description

Pagination the person's voting records

### Use Case

Currently, all of the person's voting records are initially retrieved. We can reduce the "time to interactivity" of the person page by first showing only N voting records and then allowing the user to show more voting records.

### Solution

Create some sort of `show more` UI to allow users to show more voting records.

### Alternatives

_Please describe any alternatives you've considered, even if you've dismissed them._

### Screenshots/References

_In the case of a visually oriented feature or component request, providing drawings, screenshots, and/or links to examples of what you're seeking will be very helpful._

tohuynh:
### Link to Relevant Issue

This pull request resolves #156 

### Description of Changes

This PR is also a refactor of the `useFetchEvents` hook. I generalize it to `useFetchModels` so that it can be used for fetching/paginating renderable events, votes, matter sponsored, or any other javascript object.

`useFetchModels` custom hook returns the current `state` (contains data objects and other fields to handle fetch and pagination logic) and a `dispatch` method. The custom hook uses two other React hooks -- useReducer and useEffect. 

useReducer takes an initial `state` and reducer of type `(state, action) => newState` returns the current `state` and a `dispatch`  method. It is used to handle synchronous actions -- update the `state` on success or failed data fetching/data pagination, update whether there is a request to fetch data or paginate data. useEffect is used to handle asynchronous actions -- actually fetching data. It listens to whether there has been a data request and then fetches data.

`useFetchModels` is designed to fetch any javascript object -- renderable events (in /events page), votes and matter sponsor (in /people/:id page) because it's generic.  If it follows the "show more" flow of pagination, this hook can be used. `useFetchModels` accepts two args -- the initial `state` and a `fetchModelsFunctionCreator` of type `(state) => async () => Promise<Model[]>`. The function creator creates an async fetch data function from the current `state`. In the useEffect section of `useFetchModels` the fetch function is created and then called with `const models = await fetchModelsFunctionCreator(state)();`

---
Just a heads up regarding allowing users to sort and filter the list of votes. Currently, we can only filter/sort on non-reference fields (and have the sorting/filtering make sense). We can filter `vote.decision` and filter/sort by `vote.datetime` (when it is added). Sorting/filtering by other columns would force us to fetch all votes and populate the reference fields, and this would lead us back to the bug of firestore can't handle too many requests.

https://tohuynh.github.io/cdp-frontend/#/people/e38abb6f86b2


JacksonMaxfield (approved):
Wow. Really nice work generalizing the Events page pagination to this function as well.

I think I am also noticing a speed up on page load? Somehow? I thought all of that was async loading but :woman_shrugging: maybe I am going crazy but seems like it anyway.

The grouping of the votes by event id is great too. It's not perfect but it's certainly better than before / current interaction.

JacksonMaxfield:
> Just a heads up regarding allowing users to sort and filter the list of votes. Currently, we can only filter/sort on non-reference fields (and have the sorting/filtering make sense). We can filter vote.decision and filter/sort by vote.datetime (when it is added). Sorting/filtering by other columns would force us to fetch all votes and populate the reference fields, and this would lead us back to the bug of firestore can't handle too many requests.

Yea I understand this one. I think we should still _think_ about what we can do long term. If not everything, what makes the most sense?

For example, I could see people wanting to have a search bar to filter by matter name?

tohuynh:
I like 
- search bar for searching matter name and/or top 5 keywords/topics
- filter by vote.decision (currently possible)
- filter by in majority (currently possible)
- filter by council decision (not needed, since we got in majority?)
- sort by vote date time

src/hooks/useFetchModels.ts
```diff
@@ -0,0 +1,103 @@
...
+export enum FetchModelsActionType {
+  FAILURE = "FAILURE",
+  SUCCESS = "SUCCESS",
+  FETCH_MODELS = "FETCH_MODELS",
+}
+
+export type FetchModelsAction<Model> =
+  | { type: FetchModelsActionType.FAILURE; payload: Error }
+  | { type: FetchModelsActionType.SUCCESS; payload: Model[] }
+  //The payload is whether to fetch models with different filter parameters or fetch more models with same filter parameters.
```
BrianL3, line 14: I don't really understand this comment. The payload is a boolean, so... if it changes between true/false that indicates it needs to re-fetch?

JacksonMaxfield:
> I like
> 
>     * search bar for searching matter name and/or top 5 keywords/topics
> 
>     * filter by vote.decision (currently possible)
> 
>     * filter by in majority (currently possible)
> 
>     * filter by council decision (not needed, since we got in majority?)
> 
>     * filter and sort by vote date time

Oh yes, once we have matter indexing, we can search by key terms too. Good idea!

src/hooks/useFetchModels.ts
tohuynh: action = {type: "fetch_models", payload: true} means that it's request to fetch data -- i.e. replace `models` with the new list of `models` that is incoming.

action = {type: "fetch_models", payload: false} means it's a request to fetch more data -- i.e. append the incoming list of `models` to current list of `models`.

Basically, the first action is used when user did some filtering and this necessitate fetching data. The second action is using clicking "show more".

tohuynh: action = {type: "fetch_models", payload: true} changes the state.fetchModels to true and state.fetchMoreModels to false

action = {type: "fetch_models", payload: false} changes state.fetchModels to false and state.fetchMoreModels to true

in the useEffect section, whenever state.fetchModels or state.fetchMoreModels changes, it will fetch data. Then it will dispatch an action of type `success` with payload=fetched models. Then depending on the value of state.fetchModels it replaces the models or append to it. https://github.com/tohuynh/cdp-frontend/blob/feature/paginate-votes/src/hooks/useFetchModels.ts#L34

BrianL3 (approved):

Issue #164, opened 2022-01-07
Design: The People Page


BrianL3:
### The People Page

The People Page needs two things:

1. a design
2. someone to implement that design

### Use Case

Part of this story is to have the design team come up with the different use cases of the People Page. Remember that we have a number of different user archetypes: the journalist, the academic researcher, and the layperson (average citizen). 
Is a citizen coming to this page because they do not know who their current council members are?
Does the user want to look for a specific person?
Which people are most important to show?
Are there any other considerations for searching, browsing, or perhaps "custom" looks ("show me who was on the council in XYZ year")?

The primary ways we have of looking at people are Role, time and geography.

- For more on what roles are available, talk to @JacksonMaxfield .
- Some Roles have Seats (e.g. if you are the Councilperson of the Beacon Hill neighborhood, your Seat would be there). So there is an aspect of geography available for some roles. 
- Time could be an interesting aspect, as CDP does have use as a research tool.

### Solution

The PersonCard (see the storybook library to see this component) was designed to be used on this page. 

### Alternatives

The current PeoplePage as shown is an alternative, I suppose.  It is just a list of Person links by name. I think we can all agree that it needs a re-vamp.

### Current Look Reference
![Screen Shot 2021-09-24 at 4 57 32 PM](https://user-images.githubusercontent.com/9027758/148597802-ab1705d1-c0a7-4921-9bad-355c8f9dc25e.png)

tohuynh:
@BrianL3 Is this for the `/people` page, not the `/people/:id` page right?

BrianL3:
Yes, this is for the /people page.  Sorry, I uploaded the wrong thing.

PR #209, opened 2022-03-18
component/people-page

BrianL3:
### Link to Relevant Issue

As prep for completion of #164 I added PersonCards to the PeoplePage.

### Description of Changes

I also changed the way we query for the current council (those displayed on the People page). This seems to be more accurate and lists the current council without listing any weird entities like "No Sponsor".

PersonCard was refactored as well. Each Card now receives a role(and person) and seat as props, and then fetches its own images. 

### Link to Deployed Staged Environment
See [here](https://brianl3.github.io/cdp-frontend/#/people)  to see the new PeoplePage.

### DIFF
--- a/src/containers/PeopleContainer/PeopleContainer.tsx
+++ b/src/containers/PeopleContainer/PeopleContainer.tsx
@@ -1,35 +1,69 @@
 import React from "react";
-import { PeoplePageData } from "./types";
-import Person from "../../models/Person";
-import { Link } from "react-router-dom";
-export interface PeopleContainerProps extends PeoplePageData {
+
+import Seat from "../../models/Seat";
+import Role from "../../models/Role";
+
+import H2 from "../../components/Shared/H2";
+import { PersonCard } from "../../components/Cards/PersonCard";
+
+import styled from "@emotion/styled";
+import { useMediaQuery } from "react-responsive";
+import { screenWidths } from "../../styles/mediaBreakpoints";
+
+export interface PeopleContainerProps {
   /** Any extra info */
   searchQuery?: string;
+  /** all current councilor roles (with Person populated) */
+  roles: Role[];
+  /** the seats associated with the above councilmembers (ordered) */
+  seats: Seat[];
 }
 
-function renderLinkCard(person: Person) {
-  return (
-    <div>
-      <Link to={`/people/${person.id}`}>{person.name}</Link>
-    </div>
-  );
-}
+const PersonContainer = ({ roles, seats }: PeopleContainerProps) => {
+  const isMobile = useMediaQuery({ query: `(max-width: ${screenWidths.tablet})` });
 
-const PersonContainer = ({ currentPeople, allPeople }: PeopleContainerProps) => {
+  const CardRow = styled.div({
+    display: "flex",
+    flexDirection: isMobile ? "column" : "row",
+    gap: 8,
+    flex: 1,
+  });
+
+  const CardCell = styled.div({
+    flex: 1,
+    maxWidth: "50%",
+  });
+
   return (
     <div>
-      <h3>Currently Elected</h3>
-      {currentPeople &&
-        currentPeople.map((role) => {
+      <H2>Current Councilmembers</H2>
+      {roles.map((role, index) => {
+        if (index % 2 === 0) {
+          const role2 = roles[index + 1];
           if (role.person) {
-            return renderLinkCard(role.person);
+            return (
+              <CardRow>
+                <CardCell>
+                  <PersonCard
+                    key={`person-card-${role.person?.name}`}
+                    person={role.person}
+                    seat={seats[index]}
+                  />
+                </CardCell>
+                {role2 && role2.person && (
+                  <CardCell>
+                    <PersonCard
+                      key={`person-card-${role2.person.name}`}
+                      person={role2.person}
+                      seat={seats[index + 1]}
+                    />
+                  </CardCell>
+                )}
+              </CardRow>
+            );
           }
-        })}
-      <h3>All</h3>
-      {allPeople &&
-        allPeople.map((person) => {
-          return renderLinkCard(person);
-        })}
+        }
+      })}
     </div>
   );
 };
--- a/src/networking/RoleService.ts
+++ b/src/networking/RoleService.ts
@@ -1,4 +1,4 @@
-import { where, orderBy, doc } from "@firebase/firestore";
+import { where, orderBy, doc, limit } from "@firebase/firestore";
 
 import { NetworkService } from "./NetworkService";
 import ModelService from "./ModelService";
@@ -65,14 +65,27 @@ export default class RoleService extends ModelService {
     ) as Promise<Role[]>;
   }
 
-  async getCurrentRoles(): Promise<Role[]> {
+  async getMostRecentCouncilMemberRoleBySeat(seatId: string): Promise<Role> {
     const populatePersonRef = new Populate(COLLECTION_NAME.Person, REF_PROPERTY_NAME.RolePersonRef);
 
     const networkQueryResponse = this.networkService.getDocuments(
       COLLECTION_NAME.Role,
-      [where("end_datetime", WHERE_OPERATOR.eq, null)],
+      [
+        where(
+          REF_PROPERTY_NAME.RoleSeatRef,
+          WHERE_OPERATOR.eq,
+          doc(NetworkService.getDb(), COLLECTION_NAME.Seat, seatId)
+        ),
+        limit(1),
+        orderBy("end_datetime", ORDER_DIRECTION.desc),
+      ],
       new PopulationOptions([populatePersonRef])
     );
-    return this.createModels(networkQueryResponse, Role, `getCurrentRoles`) as Promise<Role[]>;
+    const roles = await this.createModels(
+      networkQueryResponse,
+      Role,
+      `getMostRecentCouncilMemberRoleBySeat`
+    );
+    return Promise.resolve(roles[0] as Role);
   }
 }
--- a/src/networking/SeatService.ts
+++ b/src/networking/SeatService.ts
@@ -1,0 +1,16 @@
+import ModelService from "./ModelService";
+import Seat from "../models/Seat";
+import { COLLECTION_NAME } from "./PopulationOptions";
+import { FirebaseConfig } from "../app/AppConfigContext";
+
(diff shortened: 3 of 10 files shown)


src/networking/RoleService.ts
```diff
@@ -65,14 +65,27 @@ export default class RoleService extends ModelService {
...
+      ],
       new PopulationOptions([populatePersonRef])
     );
-    return this.createModels(networkQueryResponse, Role, `getCurrentRoles`) as Promise<Role[]>;
+    const roles = await this.createModels(
+      networkQueryResponse,
+      Role,
+      `getMostRecentCouncilMemberRoleBySeat`
+    );
+    return Promise.resolve(roles[0] as Role);
```
tohuynh, line 89: This could cause a run time error. What if roles.length == 0?

I think it is better to return `roles` for each seat, then when the cards are assembled only create a card for a seat if its associated `roles`.length > 0.

src/containers/PeopleContainer/PeopleContainer.tsx
```diff
@@ -1,35 +1,69 @@
...
+    maxWidth: "50%",
+  });
 
-const PersonContainer = ({ currentPeople, allPeople }: PeopleContainerProps) => {
   return (
     <div>
-      <h3>Currently Elected</h3>
-      {currentPeople &&
-        currentPeople.map((role) => {
+      <H2>Current Councilmembers</H2>
```
tohuynh, line 39: Let's use `h1` with `mzp-u-title-sm` to make it consistent with other pages

tohuynh (commented):
Nice work!

I have two asks -- one straightforward, and the other a little complicated.

Could you use `CardsContainer` to layout the person card in `PeopleContainer`? See `EventsContainer` for an example of how to use `CardsContainer`. Using `CardsContainer` will layout the cards similar to the other pages and wraps an `a` around the card (which makes it tabable and have a cursor).

I like delegating fetching the pictures to `PersonCard`. But `FetchDataContainer` is not appropriate for this component. When the images are loading, there is a bunch of loading spinners all in the same spot and as a result, it doesn't look like a spinner anymore. The better alternative is to use `PlaceholderWrapper` (similar to the cover image in the person page).  It's a little complicated to implement and I can make a PR for it.

src/networking/RoleService.ts
```diff
@@ -65,14 +65,27 @@ export default class RoleService extends ModelService {
...
       COLLECTION_NAME.Role,
-      [where("end_datetime", WHERE_OPERATOR.eq, null)],
+      [
+        where(
+          REF_PROPERTY_NAME.RoleSeatRef,
+          WHERE_OPERATOR.eq,
+          doc(NetworkService.getDb(), COLLECTION_NAME.Seat, seatId)
+        ),
+        limit(1),
+        orderBy("end_datetime", ORDER_DIRECTION.desc),
```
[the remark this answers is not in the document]
BrianL3: I don't think `role.end_datetime` is ever null. If you sort by that property in seattle-staging, there are no documents with a null end_datetime. The current roles now have real dates. 

Sorting by a property should return documents that have that property as null. I assume that they are at the end of a list sorted in descending order, though. So, if there ARE null `end_datetime` roles, they would be at the end of this list which may be a problem.  @JacksonMaxfield care to weigh in?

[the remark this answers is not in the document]
tohuynh: Yeah, I haven't see a null end_datetime yet, but it's a possibility according to https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/database/models.py#L199 (not a required field)

src/networking/RoleService.ts
BrianL3: The page relies on having matching arrays of seats and roles. I think I will maybe try throwing an error if roles.length == 0, and catching the error in PeoplePage, and there removing the seat-without-a-role from the list of stuff passed down as props?

tohuynh: You could still have matching arrays. If `getMostRecentCouncilMemberRoleBySeat` returns an array, then `roles` in PeoplePage would be of type `Role[][]` -- each seat is associated with a Role[] array (which should have only 1 element). If it doesn't don't display the seat?

BrianL3: I did it an alternate way: returning Role or null, then stripping the 'bad' Roles from the array and splicing those indices from the seats array.  Either way it's taken care of.

BrianL3:
> Could you use `CardsContainer` to layout the person card in `PeopleContainer`? See `EventsContainer` for an example of how to use `CardsContainer`. Using `CardsContainer` will layout the cards similar to the other pages and wraps an `a` around the card (which makes it tabable and have a cursor).

Did this.


> I like delegating fetching the pictures to `PersonCard`. But `FetchDataContainer` is not appropriate for this component. When the images are loading, there is a bunch of loading spinners all in the same spot and as a result, it doesn't look like a spinner anymore. The better alternative is to use `PlaceholderWrapper` (similar to the cover image in the person page). It's a little complicated to implement and I can make a PR for it.

I looked at the PlaceholderWrapper thing and was a bit confused, you can PR it if you want and I can learn how it's done.

tohuynh:
> the PlaceholderWrapper thing and was a bit confused, you can PR it if you want and I can learn how it's done.

Will do

src/networking/SeatService.ts
```diff
@@ -0,0 +1,16 @@
...
+import { COLLECTION_NAME } from "./PopulationOptions";
+import { FirebaseConfig } from "../app/AppConfigContext";
+
+export default class SeatService extends ModelService {
+  constructor(firebaseConfig: FirebaseConfig) {
+    super(COLLECTION_NAME.Seat, firebaseConfig);
+  }
+
+  async getAllSeats(): Promise<Seat[]> {
+    const networkResponse = this.networkService.getDocuments(COLLECTION_NAME.Seat, []);
```
tohuynh, line 12: Can we sort by `seat.name`?

src/networking/RoleService.ts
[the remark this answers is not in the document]
BrianL3: I think this is currently the most accurate query we have to get the full council. Maybe we correct if its starts returning accurate results? When did the `end_datetime` get populated, if it used to be `null`?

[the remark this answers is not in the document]
tohuynh: >When did the end_datetime get populated, if it used to be null?


I think for all the instances we have, end_datetime always had a date.

I think Jackson allowed for "null" end_datetime because there is a possibility of an ongoing role with no end_datetime?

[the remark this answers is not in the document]
JacksonMaxfield: > I think for all the instances we have, end_datetime always had a date.
> 
> I think Jackson allowed for "null" end_datetime because there is a possibility of an ongoing role with no end_datetime?

Yep. This is true. It may suck but is it possible to do two queries and set-addition the returns?

One query for most recent role by end datetime (like you currently have) and one query for roles with no end_datetime and then set addition them?

JacksonMaxfield (approved):
This looks good to me. Approving for now because its already a massive improvement and looks GREAT.

Some extra comments / nitpicks:

1. can we make it a three-wide card layout instead of two? Both to match the events page and because the cards look REALLY big on my screen haha
2. the page linked only shows the current councilmembers, can we get an "all councilmembers" list / table below or something? Maybe make another issue for that? Iirc, the full councilmembers table / list / w.e. was what Dale was working on and had a design for. We had some cool filter ideas. But that may be a larger issue.
3. is it possible to order the councilmembers by name or actually imo by seat name? Like it would be cool if they were always in order of "Position 1", "Position 2", "Position 3", etc.

src/networking/RoleService.ts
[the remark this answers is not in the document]
BrianL3: My only concern is that the queries for roles with null `end_datetime` does leave us in a slightly uncertain state: we are inferring that those null roles found are the most recent, but we are not certain.

I do like the set-addition query better than our currently-deployed solution though, as it is based off of seats (so we won't NO SPONSOR entities). I will implement it.

BrianL3:
can we make it a three-wide card layout 

> instead of two? Both to match the events page and because the cards look REALLY big on my screen haha
Yes, I changed the component to re-used the Cards thing from Events. I didn't know that component existed, I am quite pleased to use the generalized components.

> the page linked only shows the current councilmembers, can we get an "all councilmembers" list / table below or something? Maybe make another issue for that? Iirc, the full councilmembers table / list / w.e. was what Dale was working on and had a design for. We had some cool filter ideas. But that may be a larger issue.

I will put that and search stuff into a different PR that concentrates more on adding search.

> is it possible to order the councilmembers by name or actually imo by seat name? Like it would be cool if they were always in order of "Position 1", "Position 2", "Position 3", etc.
I may need you to make another index for this, I am not sure. I put an `orderBy` seat name, that should hopefully do it.

JacksonMaxfield:
> > is it possible to order the councilmembers by name or actually imo by seat name? Like it would be cool if they were always in order of "Position 1", "Position 2", "Position 3", etc.
> 
> I may need you to make another index for this, I am not sure. I put an `orderBy` seat name, that should hopefully do it.

Cool. We can do that in a later PR too. No worries for this one.

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -26,7 +26,26 @@ const PeoplePage: FC = () => {
...
+    const goodRoles: Role[] = roles.reduce((filtered, optional) => {
+      if (optional !== null) {
+        filtered.push(optional);
+      }
+      return filtered;
+    }, [] as Role[]);
+    for (let i = badRoles.length - 1; i >= 0; i--) {
+      // remove the bad, no-Role-having Seats
+      seats.splice(badRoles[i], 1);
+    }
```
tohuynh, line 46: Couldn't this be done in one loop?
```
roles.reduce((goodRolesAndSeats, role, index) => {
  if (role) {
     goodRolesAndSeats.roles.push(role)
     goodRolesAndSeats.seats.push(seats[index])
  }
  return goodRolesAndSeats
}, {roles: [], seats: []})
```
might have to add some typing though

tohuynh (approved):
Looks good! Thanks for making the changes.

I have maybe one tiny improvement.

JacksonMaxfield:
@BrianL3 seattle-staging index change now deployed. The new page looks great! Feel free to merge whenever.

Issue #243, opened 2022-10-19
Search transcript field should ignore case during search


evamaxfield:
https://www.openmontana.org/missoula-council-data-project/#/events/3ba34090b674

Reported by Mark from DemocracyLab.

If you type in "Abortion" in the "Search transcript" nothing is returned. If you type in "abortion" in the "Search transcript" it properly returns 9 results.

It looks like we are currently calling lowercase on the string but maybe we need to move that lowercasing to before the stemming? https://github.com/CouncilDataProject/cdp-frontend/blob/main/src/components/Details/TranscriptSearch/TranscriptSearch.tsx#L90

jonlin14:
Yes it looks like the lowercasing needs to happen before the stemming.
<img width="529" alt="Screen Shot 2023-02-14 at 10 26 30 AM" src="https://user-images.githubusercontent.com/10490944/218836019-0f36bc79-f7b7-4294-a54b-26beddb1bf5c.png">
abortion gets stemmed to abort here

<br><br><br>

<img width="451" alt="Screen Shot 2023-02-14 at 10 25 00 AM" src="https://user-images.githubusercontent.com/10490944/218836251-a0a62961-88d1-47e2-a03f-907f7a5c491f.png">

but if we have a upperCase "Abortion", the stemr library does not stem it. Then we lowerCase "Abortion" into "abortion" and now we're searching with the term "abortion" which we won't find because the stemmedSentences only has "abort".

At first I thought this was pronoun protection. So that if somebody were to search "Billings" looking for information about the city in Montana, they wouldn't get results about billing, but "Billings" does get stemmed into "bill" so I think it's just some upperCase words are just blind spots? If we lowerCase them first, this should be resolved.

PR #254, opened 2023-02-14
bug-fix/case-search-issues

jonlin14:
### Link to Relevant Issue

https://github.com/CouncilDataProject/cdp-frontend/issues/243
(I left additional notes here)

### Description of Changes

Changing transcriptSearch to lowerCase first before stemming.
Adding lowerCasing to TranscriptItem component so the words that the search results returned will now also be highlighted (for example Abortion will now highlight abort).

https://councildataproject.org/seattle/#/events/e711bed20cda
Abortion will return 0 results, abortion will return 1.

https://jonlin14.github.io/cdp-frontend/#/events/e711bed20cda
Abortion will return the expected 1 result, abortion will return 1.

### DIFF
--- a/src/components/Details/TranscriptItem/TranscriptItem.tsx
+++ b/src/components/Details/TranscriptItem/TranscriptItem.tsx
@@ -154,7 +154,7 @@ const TranscriptItem: FC<TranscriptItemProps> = ({
       // no query or valid tokens to highlight
       return [];
     }
-    const stemmedQuery = tokenizedQuery.map((token) => stem(token));
+    const stemmedQuery = tokenizedQuery.map((token) => stem(token.toLowerCase()));
     // highlight the token or the stem
     const regExps = tokenizedQuery.map(
       (token, i) => new RegExp(`\\b(${token}|${stemmedQuery[i]})`, "g")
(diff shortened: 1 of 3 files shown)


tohuynh:
This looks good!

Non-blocking: @evamaxfield, wanna switch [stem-lowerCase](https://github.com/CouncilDataProject/cdp-frontend/blob/main/src/networking/EventSearchService.ts#L99) here as well? Does cdp-backend also lowerCase first, and then stem?

evamaxfield:
> This looks good!
> 
> Non-blocking: @evamaxfield, wanna switch [stem-lowerCase](https://github.com/CouncilDataProject/cdp-frontend/blob/main/src/networking/EventSearchService.ts#L99) here as well? Does cdp-backend also lowerCase first, and then stem?

Wow. Nice catch @tohuynh. We lowercase the gram prior to stemming: https://github.com/CouncilDataProject/cdp-backend/blob/main/cdp_backend/pipeline/generate_event_index_pipeline.py#L235

I don't think search will change all that much but @jonlin14 can you make that change that @tohuynh suggested.

Otherwise all looks good to me.

jonlin14:
Made the change in getStemmedGrams to lowerCase first as well.

https://councildataproject.org/seattle/#/events
Searching for "Abortion" will return 0, "abortion" will return 23 

https://jonlin14.github.io/cdp-frontend/#/events
Searching for "Abortion" will return 1, "abortion" will return 1

evamaxfield (approved):
Looks good to me!

PR #262, opened 2023-03-12
Adds a copy button to all transcript items

alonr619:
Added a copy button to TranscriptItem.tsx.

For the svg, I used [this](https://www.svgrepo.com/svg/3110/copy) since I couldn't find an existing one in our codebase. It should be fairly simple to change it, under components/Shared/CopyIcon.tsx.

Also, I used Google Translate to translate "Copy Text" to Spanish and Dutch. If this is wrong or inaccurate, it should be simple to change under assets/strings.


src/assets/strings/de.ts
```diff
@@ -97,6 +97,7 @@ const de = {
   disclaimer_footer_start: "In vielen Fällen verwendet Council Data Project ein fein abgestimmtes",
   disclaimer_footer_end:
     "Modell zur Generierung von Ereignisprotokollen. Wir verstehen, dass Transkripte Fehler enthalten können. Wenn Sie ein Experte für maschinelles Lernen sind und dabei helfen möchten, unser System zum Generieren von Transkripten zu verbessern, kontaktieren Sie uns bitte auf GitHub.",
+  copy_button: "Kopieer Tekst",
```
evamaxfield, line 100: Sadly we still don't make use of these strings but regardless this isn't actually Dutch, it's German!

The correct string should be: "Text kopieren" -- at least by google translate results

evamaxfield (commented):
I think this all makes sense! Can you deploy this PR to your GitHub Pages.

Should be possible with `npm run deploy:app`. Would like to see what it looks like in the app prior to merging.

src/assets/strings/de.ts
BrianL3: genau 👍

evamaxfield:
Looks great! Thanks @alonr619 -- will merge after tests pass

evamaxfield:
Deployment here: https://alonr619.github.io/cdp-frontend/#/events/fbdb5ca37758

PR #265, opened 2023-04-11
New search functionality

### LINKED ISSUE #138: Implement strict search and other useful search operators
I would love this for the general event search too so I am going to work on that problem right now.

_Originally posted by @JacksonMaxfield in https://github.com/CouncilDataProject/cdp-frontend/issues/137#issuecomment-986056104_

> I would love this for the general event search too so I am going to work on that problem right now.

Pausing for a bit on this because I think alg is somewhat hard.... in psuedocode it's something like:
```
query = 'hello world testing "strict part a" "strict part b'
// some func to get to query parts
queryParts = ["hello", "world", "testing", '"strict part a"', '"strict part b"']
// some func to move to ngrams for non strict parts
ngramSearches = ["hello", "hello world", "hello world testing", "world", "world testing", "testing"]
// some func to move to just strict parts
strictSearches = ["strict part a", "strict part b"]
// some func to clean and stem both
cleanedNGramSearches = ["hello", "hello world", "hello world test", "world", "world test", "test"]
cleanedStrictSearches = ["strict part a", "strict part b"]
// run queries for the strict searches first
// run queries for the grams like how we are currently doing it but filtered by events that matched the strict results first??? (which means really all these results are doing are adding weighting to events which is fine)
// merge to unified event gram results
```

But that leaves interesting questions like how do we handle AND and OR. Need to look to google for how they are doing it basically imo.

alonr619:
…ltering.

### Link to Relevant Issue

This pull request resolves #138 

### Description of Changes

Adds new search functionality to event transcript search, including strict search (use double quotes), negative search (put a ! before the double quotes), start filtering (put a ^ before the double quotes) and end filtering (put a $ before the double quotes).

For example, `^"1" $"2" "3" !"4"` would only show strings that start with 1, end with 2, include 3, and don't include 4.

### Linked to Forked Storybook Site

[Here](https://alonr619.github.io/cdp-frontend/#/events/b4e8a57e226d)

### DIFF
--- a/src/components/Details/TranscriptSearch/TranscriptSearch.tsx
+++ b/src/components/Details/TranscriptSearch/TranscriptSearch.tsx
@@ -12,6 +12,7 @@ import { strings } from "../../../assets/LocalizedStrings";
 import { fontSizes } from "../../../styles/fonts";
 import { screenWidths } from "../../../styles/mediaBreakpoints";
 import cleanText from "../../../utils/cleanText";
+import { varietyRoles } from "../../../stories/model-mocks/role";
 
 const Container = styled.div({
   display: "flex",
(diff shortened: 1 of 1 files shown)


jonlin14 (commented):
On first pass 
- linter is saying varietyRoles on line 15 isn't used, is it necessary? If not lets remove it.
- Lets also get rid of the console logs.
- I tried testing "Thank" !"you", looking for sentences that has only "Thank" but not "you" and it errored out, unable to read undefined from text. It looks like it's on line 188, I think if there are results from a double quote search but then they all subsequently get filtered out from using !"", it's trying to read from an answer with no results thus erroring out.
- Filtering seems to only be working for full strings right now, is that intended? Also looks like the highlighting is off. It can be adjusted in the TranscriptItem.tsx file. You can look at line 177 to see the Highlighter component.
filter1

filter2


@evamaxfield 
Should we include text that tells the user that they can use quotes and filters?
Should this be case sensitive?

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
```diff
@@ -12,6 +12,7 @@ import { strings } from "../../../assets/LocalizedStrings";
 import { fontSizes } from "../../../styles/fonts";
 import { screenWidths } from "../../../styles/mediaBreakpoints";
 import cleanText from "../../../utils/cleanText";
+import { varietyRoles } from "../../../stories/model-mocks/role";
```
BrianL3, line 15: You'll want to clean that up, the linter says it's never used.

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
```diff
@@ -100,14 +101,97 @@ const TranscriptSearch: FC<TranscriptSearchProps> = ({
...
+        } else {
+          required.push(newTerm.substring(indices[i * 2] + 1, indices[i * 2 + 1]));
+        }
+        newTerm =
+          newTerm.substring(0, indices[i * 2]) +
+          newTerm.substring(indices[i * 2 + 1] + 1, newTerm.length);
+      }
+
+      let ans = [];
+      console.log("1");
```
BrianL3, line 148: clean out console logs

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
```diff
@@ -100,14 +101,97 @@ const TranscriptSearch: FC<TranscriptSearchProps> = ({
...
+        if (newTerm.substring(indices[i * 2] - 1, indices[i * 2]) === "!") {
+          banned.push(newTerm.substring(indices[i * 2] + 1, indices[i * 2 + 1]));
+        } else {
+          required.push(newTerm.substring(indices[i * 2] + 1, indices[i * 2 + 1]));
+        }
+        newTerm =
+          newTerm.substring(0, indices[i * 2]) +
+          newTerm.substring(indices[i * 2 + 1] + 1, newTerm.length);
+      }
+
```
BrianL3, line 146: can you pull everything from line 104-145ish into a utility function that just lives a neighboring folder? e.g. util/<function for getting restricter>, util/<function for getting double quotes>, etc?  Cleaner and more testable to keep the logic outside of views, as each function can be assessed on its sole purpose.

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
```diff
@@ -100,14 +101,97 @@ const TranscriptSearch: FC<TranscriptSearchProps> = ({
...
+      let ans = [];
+      console.log("1");
+      const cleanedQuery = cleanText(newTerm);
+      console.log("2");
       const tokenizedQuery = removeStopwords(cleanedQuery.split(" "));
+      console.log("3");
       if (!cleanedQuery || tokenizedQuery.length === 0) {
         // empty query or no valid tokens to search
```
BrianL3, line 154: it seems this comment needs to be updated?

BrianL3 (changes requested):
This is a prime candidate for tests. The logic in it is complex enough that I can't tell if it works at a glance, but you could 
1. split this chunk of code into smaller, testable chunks
2. set up some mocks with test queries (with quotes, limiters, etc)
3. test each small chunk to make sure it does what you think it does
4. test the code as a whole to make sure it works

There are some testing files in the repo already, use them to see best practices.

evamaxfield:
> @evamaxfield
> Should we include text that tells the user that they can use quotes and filters?
> Should this be case sensitive?

Yes absolutely. Please include some sort of helper text. Can be on hover. Can also save that for a future PR though.

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
alonr619: Just fixed this! It was there before I added my code, not sure how it passed the lints previously

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
alonr619: Fixed these too, sorry!

src/components/Details/TranscriptSearch/TranscriptSearch.tsx
alonr619: Will do this once I fix John's bugs, thanks for the suggestion!

PR #274, opened 2023-07-19
enhancement/expansion-icons

### LINKED ISSUE #226: enhancement/expansion-icons
### Feature Description

We need to make the VoteBar components clearly "clickable". They expand to show every vote in that bar, but it is not clear that they can be clicked.

### Solution

> I think the triangle next to Votes is a common UI pattern used to indicate something is an accordion -- that is there are some more details if clicked on. Though the inconsistency between the triangle next to Votes and the + sign next to See Documents sucks. Maybe make them both have triangles or +s?
- To

### Alternatives

The current experience has a hover mouseover experience. 

### Screenshots/References
Screen Shot 2022-06-07 at 1 34 45 PM
Screen Shot 2022-06-07 at 1 35 07 PM

BrianL3:
### Link to Relevant Issue

This pull request resolves #226 

### Description of Changes

On the legislative history page, if you see the Latest Vote section, it is not immediately apparent that you can open each Vote Bar to see who voted. I added some expansion icons to make this more clear.

[storybook](https://brianl3.github.io/cdp-frontend/?path=/story/library-details-legislation-latestvote--adopted)


evamaxfield (approved):
Good idea!!

